PR #8, opened 2015-02-19
Passwords

stuartornum:


fabfile.py
```diff
@@ -64,7 +69,10 @@ def get_config():
         sys.exit(1)
 
     aws_config = AWSConfig(env.aws)
-    project_config = ProjectConfig(env.config, env.environment)
+    if env.passwords is not None:
+        project_config = ProjectConfig(env.config, env.environment, passwords=env.passwords)
+    else:
+        project_config = ProjectConfig(env.config, env.environment)
```
kyriakosoikonomakos: This if block isn't needed since 'passwords' argument has a default value of None anyway, and the method checks for it.

I'd just use project_config = ProjectConfig(env.config, env.environment, passwords=env.passwords)

helpers/config.py
```diff
@@ -31,10 +32,34 @@ class ProjectConfig:
 
     config = None
 
-    def __init__(self, config, environment):
-        if os.path.exists(config):
-            f = yaml.load(open(config).read())
-            self.config = f[environment]
+    def __init__(self, config, environment, passwords=None):
+        self.config = self.load_yaml(config)[environment]
+        if passwords is not None:
```
kyriakosoikonomakos: I'd change this to either just 'if passwords' or 'if isinstance(passwords, basestring)'

helpers/config.py
```diff
@@ -31,10 +32,34 @@ class ProjectConfig:
...
+            passwords_dict = self.load_yaml(passwords)[environment]
+            self.config = self.dict_merge(self.config, passwords_dict)
+
+    @staticmethod
+    def load_yaml(fp):
+        if os.path.exists(fp):
+            return yaml.load(open(fp).read())
+
+    def dict_merge(self, target, *args):
+        # Merge multiple dicts
```
kyriakosoikonomakos, line 47: no tests for dict_merge?

stuartornum, line 47: Separate PR coming for tests

PR #14, opened 2015-02-24
 allow elb names with '.'

pidah:


milosgajdos:
You also need to modify the line where ELB name is assigned:

``` python
template['ElasticLoadBalancer']['Properties']['LoadBalancerName'] = 'ELB-%s' % elb['name']
```

ELB name is populated with the same value as DNS zone entry - or rather it's the other way around. Point being ELB name can not contain dots because of AWS dramas...

helpers/config.py
```diff
@@ -229,9 +223,9 @@ def elb(self):
                 'AliasTarget']['DNSName']['Fn::GetAtt'] = target_dns
 
             elb_list.append(
-                {'ELB%s' % elb['name'].replace('-', ''): template['ElasticLoadBalancer']})
+                {'ELB%s' % elb['name'].replace('-', '').replace('.', ''): template['ElasticLoadBalancer']})
             elb_list.append(
-                {'DNS%s' % elb['name'].replace('-', ''): template['DNSRecord']})
+                {'DNS%s' % elb['name'].replace('-', '').replace('.', ''): template['DNSRecord']})
```
kyriakosoikonomakos: string.replace(s, old, new[, maxreplace])
 string.replace(s, old, new[, maxreplace])¶
    Return a copy of string s with all occurrences of substring old replaced by new. If the optional argument maxreplace is given, the first maxreplace occurrences are replaced.

PR #16, opened 2015-02-26
Use pkgutil.get_data to load files, rather than looking relative to cwd

ashb:
This means that you can run this from anywhere, not just in the actual
bootstrap-cfn checkout


ashb:
I tested this in ipython but not directly using fabric - I (or someone) still needs to test that.

PR #19, opened 2015-02-26
Add tests for PR #14

ashb:


tests/tests.py
```diff
@@ -201,6 +201,68 @@ def test_elb(self):
                 'dev').config)
         self.assertEquals(known, config.elb())
 
+    def test_elb_with_reserved_chars(self):
+        from testfixtures import compare
+
+        self.maxDiff = None
```
milosgajdos, line 207: Isn't this already set globally in `TestConfigParser(unittest.TestCase). setUp ()` ?

ashb, line 207: It wasn't long enough - this case had a diff of over 5000 characters when it wasn't right.

milosgajdos:
LGTM :+1:

PR #20, opened 2015-02-27
SSL on ELBs

mattmb:
This first: #21

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,13 +98,14 @@ dev:
     - name: test-dev-external
       hosted_zone: my.domain.com.
       scheme: internet-facing
+      certificate_name: my-cert
       listeners:
         - LoadBalancerPort: 80
           InstancePort: 80
           Protocol: TCP
         - LoadBalancerPort: 443
           InstancePort: 443
-          Protocol: TCP
+          Protocol: HTTPS
     - name: test-dev-internal
       hosted_zone: my.domain.com.
       scheme: internet-facing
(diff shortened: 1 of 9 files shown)


stacks/elb_ssl.json
```diff
@@ -0,0 +1,14 @@
...
+       [
+          "arn:aws:iam::",
+          {
+             "Ref":"AWS::AccountId"
+          },
+          ":server-certificate/"
+       ]
+    ]
+  }
+}
```
ashb: This file is in the wrong place - you've updated the code to use pkgutil.get_data but not moved the file to live at awsutils/stacks/elb_ssl.json

mattmb: Good spot, thanks :+1:

README.md
```diff
@@ -126,5 +127,23 @@ dev:
...
+        blahblahblah
+        -----END RSA PRIVATE KEY-----
+      chain: |
+        -----BEGIN CERTIFICATE-----
+        blahblahblah
+        -----END CERTIFICATE-----
 ```
 
+##### SSL certs for ELBs
+If you set the protocol on an ELB to HTTPS you must include a key called `certificate_name` and matching cert data in a key with the same name under `ssl` (see example above). The `cert` and `key` are required and the `chain` is optional.
```
ashb: `certificate_name`  given the example YAML shuld be `my-cert` I believe.

We also perhaps should make it clear that we recommend storing the cert in the secrets/passwords YAML, not the main project one.

ashb: Oh actually it would be `my-cert-$stackname` wouldn't it

mattmb: let me update that so it's clearer, we just add that stack-name when we upload/delete not in the config.

tests/tests.py
```diff
@@ -198,7 +198,134 @@ def test_elb(self):
...
+                  'Protocol': 'TCP'
+                },
+                { 'LoadBalancerPort': 443,
+                  'InstancePort': 443,
+                  'Protocol': 'HTTPS'
+                },
+            ],
+        }]
+        config = ConfigParser(project_config.config,'my-stack-name')
+        self.assertRaises(errors.CfnConfigError)
```
ashb: You aren't actually testing anything here: when called like this it is returning an Context Manager: https://docs.python.org/2.7/library/unittest.html#unittest.TestCase.assertRaises

``` python
with self.assertRaises(errors.CfnConfigError):
  config.ssl()
```

is what you want

tests/tests.py
```diff
@@ -198,7 +198,134 @@ def test_elb(self):
...
```
ashb: Same here - missing the call

ashb:
Some rebasing and squashing and we should be good to :ship: :it:.

mattmb:
@ashb fixed the tests and squashed it down a bit :smile:

PR #21, opened 2015-02-27
Feature/delete stack

mattmb:
- Add ability to delete the stack
- Add a random id into each stack name, allows creating multiple stacks with the same name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,10 @@ Bootstrap-cfn uses [fabric](http://www.fabfile.org/), so if your ```$CWD``` is t
 ```
 fab application:courtfinder aws:prod environment:dev config:/path/to/courtfinder-dev.yaml cfn_create
 ```
+To delete this stack, run:
+```
+fab application:courtfinder aws:prod environment:dev config:/path/to/courtfinder-dev.yaml cfn_delete
+```
 
 If your ```CWD``` is anywhere else, you need to pass in a path to particular fabric file:
 
--- a/README.md
+++ b/README.md
@@ -35,13 +35,20 @@ fab application:courtfinder aws:prod environment:dev config:/path/to/courtfinder
 If your ```CWD``` is anywhere else, you need to pass in a path to particular fabric file:
 
 ```
-fab -f /path/to/bootstrap-cfn/fabfile.py application:courtfinder aws:prod environment:dev config:/path/to/courtfinder-dev.yaml cfn_create
+fab application:courtfinder-a1e8 aws:prod environment:dev config:/path/to/courtfinder-dev.yaml cfn_delete
 ```
+and `cfn_delete:force` to skip the prompt.
 
 * **application:courtfinder** - is just a name to associate with Cloudformation stack
 * **aws:dev** - is a way to differentiate between AWS accounts (`~/.config.yaml`)
 * **environment:dev** - The `dev` section will be read from the projects YAML file (line 1 in the example below)
 * **config:/path/to/file.yaml** - The location to the project YAML file
+
+If your ```CWD``` is anywhere else, you need to pass in a path to particular fabric file:
+
+```
+fab -f /path/to/bootstrap-cfn/fabfile.py application:courtfinder aws:prod environment:dev config:/path/to/courtfinder-dev.yaml cfn_create
+```
 
 ## Example Configuration
 ##### AWS Account Configuration
--- a/awsutils/cloudformation.py
+++ b/awsutils/cloudformation.py
@@ -24,14 +24,10 @@ class Cloudformation:
         stack = self.conn_cfn.create_stack(stack_name=stack_name,
                                            template_body=template_body,
                                            capabilities=['CAPABILITY_IAM'])
-
-        print "\n\n\n%s\n\n\nSTACK CREATED" % stack
         return stack
 
     def delete(self, stack_name):
         stack = self.conn_cfn.delete_stack(stack_name)
-
-        print "\n\n\n%s\n\n\nSTACK DELETED" % stack
         return stack
 
     def stack_done(self, stack_id):
--- a/awsutils/cloudformation.py
+++ b/awsutils/cloudformation.py
@@ -28,6 +28,12 @@ class Cloudformation:
         print "\n\n\n%s\n\n\nSTACK CREATED" % stack
         return stack
 
+    def delete(self, stack_name):
+        stack = self.conn_cfn.delete_stack(stack_name)
+
+        print "\n\n\n%s\n\n\nSTACK DELETED" % stack
+        return stack
+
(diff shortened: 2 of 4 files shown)


pidah:
Can we add a warning prompt requesting confirmation before deleting ?

mattmb:
@pidah I did think that would be nice, will take a quick look.

@kyriakosoikonomakos @ashb see https://github.com/ministryofjustice/bootstrap-cfn/commit/fd020e390e6329146f1f13e8dbf03a0174140320 does that look better?

awsutils/cloudformation.py
```diff
@@ -28,6 +28,12 @@ def create(self, stack_name, template_body):
         print "\n\n\n%s\n\n\nSTACK CREATED" % stack
         return stack
 
+    def delete(self, stack_name):
+        stack = self.conn_cfn.delete_stack(stack_name)
+
+        print "\n\n\n%s\n\n\nSTACK DELETED" % stack
```
ashb: Two things here:

1) This has only requested the stack to be deleted when this is printed
2) Info output like this shouldn't really be in this module, but in the fabfile (single responsibility principle)

awsutils/cloudformation.py
```diff
@@ -61,3 +67,8 @@ def get_stack_instances(self, stack_name_or_id):
     def get_stack_instance_ids(self, stack_name_or_id):
         return [
             x.instance_id for x in self.get_stack_instances(stack_name_or_id)]
+
+    def stack_deleted(self, stack_name_or_id):
+        ''' Returns True if stack not found'''
+        stacks = self.conn_cfn.describe_stacks()
+        return stack_name_or_id not in [s.stack_name for s in stacks]
```
ashb: The argument is called `stack_name_or_id` but it looks like it has to be an actual id. Can we make it be called what it is please?

fabfile.py
```diff
@@ -88,12 +94,40 @@ def get_config():
...
+            sys.exit(1)
+        attempts += 1
+        time.sleep(RETRY_INTERVAL)
 
 @task
 def cfn_create():
     aws_config, cfn, cfn_config = get_config()
     # Inject security groups in stack template and create stacks.
-    stack_name = "%s-%s" % (env.application, env.environment)
+    stack_name = get_new_stack_name()
```
ashb, line 128: This is a different feature to the PR at hand. Could you split that out.

Also: How does delete work if the stack name is just 'et-dev' but it's actually 'et-dev-12a5'?

fabfile.py
```diff
@@ -88,12 +94,40 @@ def get_config():
...
+        sys.exit(0)
+
+    # Wait for stacks to delete
+    print 'Waiting for stack to delete.'
+    attempts = 0
+    while True:
+        if cfn.stack_deleted(stack_name):
+           print "Stack successfully deleted"
+           break
+        if attempts == TIMEOUT / RETRY_INTERVAL:
```
ashb: What happens if you go passed the number of attempts because someone changes one of the values? `>=` please :)

awsutils/cloudformation.py
ashb: And minor point but it would be much clearer if we named this method `stack_exists` and invert the return - this meth doesn't actually have anything to do with deleting (other than how we are using it right now)

fabfile.py
```diff
@@ -88,12 +94,40 @@ def get_config():
...
+            print '[ERROR] Stack creation timed out'
```
ashb: Oh and you copy-pastad this without updating the message. We are deleting here ;)

awsutils/cloudformation.py
mattmb: yeah, it actually has to be a name now, I'll rename. I prefer stack_deleted, that way it doesn't have to be inverted in the fabfile. Happy to go with stack_missing or stack_not_exist if you prefer.

fabfile.py
mattmb: You caught me :-P

fabfile.py
mattmb, line 128: Yep, I wanted to include it because it's related in my head. I've put the 12a5 in the middle. Like et-12a5-dev, so you can do a delete with application:et-12a5 (see the README for e.g.)

mattmb, line 128: It's not perfect, maybe we could do something more magic?

ashb, line 128: There's already too much magic in this tool ;) Lets keep it simple.

How do we go from 'I asked for env:prod application:courtfinder' to 'the stack I want to delete is called courtfinder-a1e8'?

fabfile.py
```diff
@@ -6,9 +6,11 @@
 import random
 import yaml
 import time
+import uuid
 from helpers.config import AWSConfig, ProjectConfig, ConfigParser
 from awsutils.cloudformation import Cloudformation
 from awsutils.ec2 import EC2
+from boto.exception import BotoServerError
```
ashb: Unused import now.

README.md
```diff
@@ -27,18 +27,23 @@ Bootstrap-cfn uses [fabric](http://www.fabfile.org/), so if your ```$CWD``` is t
...
 ```
-
-If your ```CWD``` is anywhere else, you need to pass in a path to particular fabric file:
-
+This will create a stack with a name like `courtfinder-a1e8-prod`. To delete this stack, run:
 ```
-fab -f /path/to/bootstrap-cfn/fabfile.py application:courtfinder aws:prod environment:dev config:/path/to/courtfinder-dev.yaml cfn_create
+fab application:courtfinder-a1e8 aws:prod environment:dev config:/path/to/courtfinder-dev.yaml cfn_delete
 ```
+and `cfn_delete:force` to skip the prompt.
```
ashb, line 40: It's worth knowing that this actually ends up with calling `cfn_delete(force="force")`.

I'm not sure it matters in this case but if we want it can be invoked like `fab cfn_delete:force=True`

ashb:
(Travis tests have failed because there's no .travis.yml in this branch and travis defaults to making it a ruby project. I have changed the settings to only build if .travis.yml is present)

awsutils/cloudformation.py
```diff
@@ -61,3 +63,8 @@ def get_stack_instances(self, stack_name_or_id):
     def get_stack_instance_ids(self, stack_name_or_id):
         return [
             x.instance_id for x in self.get_stack_instances(stack_name_or_id)]
+
+    def stack_missing(self, stack_name_or_id):
```
ashb: @mattmb argument name is still `stack_name_or_id`

ashb:
Okay, one small change that still needs doing ('stack_name_or_id' to 'stack_name') and then it'll need rebasing down into.. (probably) 1 commit. It's only +98/-9  in the end after all our back and forth

PR #22, opened 2015-03-02
Fix classes

pidah:


ashb:
Chatted with Peter and we're going to do this in a different way.

PR #25, opened 2015-03-04
Fix getting stack name.

mattmb:
@pidah fixes the stack_name bug I introduced.

### DIFF
--- a/fabfile.py
+++ b/fabfile.py
@@ -64,9 +64,13 @@ def blocking(x):
 def user(x):
     env.user = x
 
-def get_new_stack_name():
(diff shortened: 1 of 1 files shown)


pidah:
rsync breaks now on find_master() – might need to update:
 @task
 def find_master():
      try:
          stack_name = env.stack_name
      except AttributeError:
          stack_name = '%s-%s' % (env.application, env.environment)
          env.stack_name = stack_name

mattmb:
@pidah good point! I've tidied it up a bit more.

pidah:
:+1:

PR #27, opened 2015-03-05
Bug Fixes

kyriakosoikonomakos:
- Filter instances per stack when looking for master
- env.passwords is used internally in fabric.
- get_new_stack_name renamed to get_stack_name
- code requires SSL cert details even when SSL isn't used.

### DIFF
--- a/bootstrap_cfn/ec2.py
+++ b/bootstrap_cfn/ec2.py
@@ -1,6 +1,6 @@
 import sys
 import boto.ec2
-
+from bootstrap_cfn import cloudformation
 
 class EC2:
 
@@ -17,6 +17,7 @@ class EC2:
         else:
             print "[ERROR] No AWS credentials"
             sys.exit(1)
+        self.cfn = cloudformation.Cloudformation(config)
 
     def get_instance_public_ips(self, instance_id_list):
         if not instance_id_list:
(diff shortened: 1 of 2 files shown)


fabfile.py
```diff
@@ -86,13 +83,15 @@ def get_config():
...
+        env.stack_passwords={}
     aws_config = AWSConfig(env.aws)
     project_config = ProjectConfig(
         env.config,
         env.environment,
-        passwords=env.passwords)
+        passwords=env.stack_passwords)
 
-    cfn_config = ConfigParser(project_config.config, env.stack_name)
+    cfn_config = ConfigParser(project_config.config, get_stack_name)
```
mattmb: shouldn't that be get_stack_name()

fabfile.py
```diff
@@ -270,7 +275,7 @@ def install_master():
     stack_ips.remove(master_prv_ip)
     stack_public_ips = ec2.get_instance_public_ips(instance_ids)
     stack_public_ips.remove(master_public_ip)
-
+    print master_public_ip
```
ashb: Looks like a debug print left in here

bootstrap_cfn/ec2.py
```diff
@@ -1,6 +1,6 @@
 import sys
 import boto.ec2
-
+from awsutils import cloudformation
```
ashb: This should be `from bootstrap_cfn import cloudformation`

kyriakosoikonomakos:
The test that failed needs to be fixed... the failure is on whitespace.

PR #33, opened 2015-03-12
Wait for SSH

mattmb:
This adds a function to wait for ssh on all instances. This is hooked into the bootstrap commands so that we only bootstrap once we can connect with ssh on all instances in the stack.


mattmb:
@pidah good shout! Have updated.

pidah:
:+1:

PR #34, opened 2015-03-13
updated fabfile to check that ssl cert exists before it tries to delete ...

pwyborn:
@mattmb Hi Matt please can you check the changes as per our discussion yesterday (on roll back check if ssl cert exists before trying to delete it"

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
  * Rename fabfile to fab_tasks to make it easier to import in other projects.
  * Move bootstrap script execution to fabric tasks.
  * Fix bug in wait_for_ssh when no instances are running.
+ * Add conditional statement in fabfile to check for ssl cert on roll back before trying to delete it.
 
 ## Version 0.1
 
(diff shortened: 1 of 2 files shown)


fabfile.py
```diff
@@ -148,9 +148,9 @@ def cfn_create():
         print 'Successfully built stack {0}.'.format(stack)
     else:
         print 'Failed to create stack: {0}'.format(stack)
-        #So delete the SSL cert that we uploaded
-        iam.delete_ssl_certificate(cfn_config.ssl(), stack_name)
-
+        #So delete the SSL cert that we uploaded (if it exists)
+        if 'ssl' in cfn_config.data:
+        	iam.delete_ssl_certificate(cfn_config.ssl(), stack_name)
```
mattmb: Two things. It also needs to create the iam object, should be like the create above. Also the indentation is not quite right.

bootstrap_cfn/fabfile.py
```diff
@@ -0,0 +1,387 @@
...
+        remote_dir=remote_state_dir,
+        local_dir=local_salt_dir,
+        use_sudo=True)
+    upload_project(
+        remote_dir=remote_pillar_dir,
+        local_dir=os.path.join(
+            local_pillar_dir,
+            env.environment,
+            '.'),
+        use_sudo=True)
```
mattmb: I think we do need a symlink here but we don't want to commit the whole file again :-)

mattmb:
@pwyborn the actual fix looks good now :+1:  Let's rebase this against master and squash it down tomorrow morning.

PR #37, opened 2015-03-16
Fix bootstrap ordering

mattmb:
- Ensure we wait for at least one instance before we bootstrap
- Run bootstrap from one place (fabric task)

(my thinking here is that anything that needs to be synchronous should run in the bootstrap task, LaunchConfiguration should only be used when we don't need to know the result)


pidah:
LGTM

PR #38, opened 2015-03-18
Add a script for running salt

mattmb:
This adds a salt_utils module which can run salt on the master, gather and parse the results to decide if state execution succeeded. Otherwise running a salt state always produces a 0 error code.

I've deliberately not included salt as a requirement seeing as this will only run on the remote machines where salt will already be installed.

### DIFF
--- a/bootstrap_cfn/salt_utils.py
+++ b/bootstrap_cfn/salt_utils.py
@@ -1,0 +1,49 @@
+#!/usr/bin/env python
+from bootstrap_cfn import utils
+from bootstrap_cfn import errors
+import salt
+import salt.runner
+import salt.client
+import pprint
+import time
+import sys
+
+def start_highstate(target):
+    local = salt.client.LocalClient()
+    jid = local.cmd_async(target, 'state.highstate')
+    return jid
+
+def start_state(target, state):
+    local = salt.client.LocalClient()
+    jid = local.cmd_async(target, 'state.sls', [state])
+    return jid
+
+def state_result(jid):
+    opts = salt.config.master_config('/etc/salt/master')
+    r = salt.runner.RunnerClient(opts)
+    result = r.cmd('jobs.lookup_jid', [jid])
+    if result:
+        return result
+    return False
+
+def highstate(target, timeout, interval):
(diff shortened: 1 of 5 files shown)


bootstrap_cfn/salt_utils.py
```diff
@@ -0,0 +1,49 @@
...
+    jid = local.cmd_async(target, 'state.sls', [state])
+    return jid
+
+def state_result(jid):
+    opts = salt.config.master_config('/etc/salt/master')
+    r = salt.runner.RunnerClient(opts)
+    result = r.cmd('jobs.lookup_jid', [jid])
+    if result:
+        return result
+    return False
```
kyriakosoikonomakos, line 27: This could be writte as "return result if result else False" and get rid lines 25, 26

bootstrap_cfn/salt_utils.py
```diff
@@ -0,0 +1,49 @@
...
+    return check_state_result(res)
+
+def state(target, state, timeout, interval):
+    jid = start_state(target, state)
+    res = utils.timeout(timeout, interval)(state_result)(jid)
+    return check_state_result(res)
+
+def check_state_result(result):
+    results = []
+    for minon, ret in result.items():
```
kyriakosoikonomakos: typo, you probably meant minion

mattmb: I did and I'm not using it anyway so result.values() would be better

kyriakosoikonomakos:
I don't see where is this called ?

bootstrap_cfn/salt_utils.py
kyriakosoikonomakos: Yes.

bootstrap_cfn/salt_utils.py
stuartornum, line 27: it's easier to read as it is

kyriakosoikonomakos, line 27: just saying :)

mattmb:
@kyriakosoikonomakos see https://github.com/ministryofjustice/template-deploy/issues/2

PR #40, opened 2015-03-20
RDS Output

mattmb:
Add a cloudformation output for the RDS instance. This makes it possible for instances in the stack to query the CloudFormation API to get the hostname and port of the RDS instance.


bootstrap_cfn/stacks/rds_out.json
```diff
@@ -0,0 +1,10 @@
+{
```
stuartornum, line 1: whats this file for?

bootstrap_cfn/stacks/iam.json
```diff
@@ -16,6 +16,7 @@
                     { "Action": [ "ec2:Describe*" ], "Resource": "*", "Effect": "Allow" },
                     { "Action": [ "rds:Describe*" ], "Resource": "*", "Effect": "Allow" },
```
stuartornum, line 17: this line is where you can query the API from relevant RDS' - although makes sense to add cloudformation API as well

ashb:
I chatted to Stuart before he had to leave and there was a bit of a misunderstanding about where the output goes (they don't go to disk - just to the Cfn API) so he's happy with it now.

PR #41, opened 2015-03-20
Fix master bootstrap

mattmb:
Now that we create in a VPC the SG for the master/minion comms needs to be in a VPC. This ensures the SG is based on the masters VPC rather than the default.


bootstrap_cfn/ec2.py
```diff
@@ -36,9 +36,13 @@ def get_instance_private_ips(self, instance_id_list):
...
 
-    def create_sg(self, name):
+    def create_sg(self, name, vpc_id):
         return self.conn_ec2.create_security_group(
-            name, 'bootstrap generated SG')
+            name, 'bootstrap generated SG', vpc_id)
+
+    def delete_sg(self, name):
+        return self.conn_ec2.delete_security_group(
+            name, 'bootstrap deleted SG')
```
ashb: What is this string literal parameter?

mattmb: Think it goes to the description on the SG.

ashb: Two further questions:

1) Why does delete have one too?
2) Is it possible to have a better indicator of what the SG actually is/does/what purpose it serves?

mattmb:
Turns out we have a problem. I can't find a way to delete a SG in a non default VPC when we only have its name. See https://github.com/boto/boto/issues/2651

ashb:
Does this mean it's time to do it properly and create the SG as part of the CFN stack?

mattmb:
It does and I have :+1:

bootstrap_cfn/fab_tasks.py
```diff
@@ -371,3 +356,15 @@ def rsync():
...
+            remote_pillar_dir,
+            'cloudformation.sls'),
+        local_path=env.config,
+        use_sudo=True)
+    put(
+        remote_path=os.path.join(
+            remote_pillar_dir,
+            'secrets.sls'),
+        local_path=env.stack_passwords,
+        use_sudo=True)
```
mattmb: Ugh, didn't mean to commit this yet...

bootstrap_cfn/stacks/ec2.json
```diff
@@ -8,11 +8,26 @@
...
+  "SaltSG": {
+    "Type": "AWS::EC2::SecurityGroup",
+    "Properties": {
+        "VpcId" : { "Ref" : "VPC" },
+        "GroupDescription": "Salt Comms Security Group",
+        "SecurityGroupIngress": [
+                                    {
+                                        "IpProtocol": "tcp",
+                                        "FromPort": 4505,
+                                        "ToPort": 4506,
```
ashb, line 20: Just to check: is this the right way to define the port range, or does it map 4505 to 4506?

stuartornum, line 20: Good spot - I think this should be fromport: 4505, toport:4505 - from being 0.0.0.0/0:4505 to somewhere in the VPC on port 4505

stuartornum, line 20: 0.0.0.0/0 is an example - it's probably 10.0.0.0/16 - which is:

{ "Fn::FindInMap" : [ "SubnetConfig", "VPC", "CIDR" ]}

stuartornum, line 20: Actually, do we need to hardcode these SG into the stack, can't it go in the YAML config file?

mattmb, line 20: Think it might be okay as is.

See [here](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-security-group-ingress.html)

What I intend it to be is open ports 4505 and 4506 inbound to anyone in BaseHostSG

mattmb, line 20: @stuartornum that's a good question _checking_

mattmb, line 20: So... the difficulty is with circular references. The `security_groups` key in the yaml just gets dumped into BaseHostSG and `SourceSecurityGroupId` can't reference itself. So the answer is it can't go in the yaml without updating the code to create security groups rather than security group rules.

stuartornum:
Spoke with @mattmb offline about how to solve this another way without the need to hardcode the SG - another PR in the pipeline...

bootstrap_cfn/stacks/ec2.json
ashb, line 20: I think that the security_groups being dumped raw into the JSON is starting to cause more problems than it solves.

PR #45, opened 2015-03-23
Add CF config to salt pillar.

mattmb:
This uploads the stack config to cloudformtaion.sls in the salt pillar dir. This way it can be included in the pillar in the deploy repositories and referenced in the salt templates.

### DIFF
--- a/bootstrap_cfn/fab_tasks.py
+++ b/bootstrap_cfn/fab_tasks.py
@@ -2,14 +2,16 @@
 
 import sys
 import random
+import yaml
 from bootstrap_cfn.config import AWSConfig, ProjectConfig, ConfigParser
 from bootstrap_cfn.cloudformation import Cloudformation
 from bootstrap_cfn.ec2 import EC2
 from bootstrap_cfn.iam import IAM
 
 import os
+from StringIO import StringIO
 
-from fabric.api import env, task, sudo
+from fabric.api import env, task, sudo, put
 from fabric.contrib.project import upload_project
 
 # GLOBAL VARIABLES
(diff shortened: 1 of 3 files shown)


bootstrap_cfn/fab_tasks.py
```diff
@@ -93,7 +93,7 @@ def get_config():
 
     cfn_config = ConfigParser(project_config.config, get_stack_name())
     cfn = Cloudformation(aws_config)
-    return aws_config, cfn, cfn_config
+    return aws_config, cfn, cfn_config, project_config
```
ashb: Hmmmm this is fast going beyond the point of readability/maintainability.

We need to come up with a better solution to this.

stuartornum:
Can you update the example config files and README please

bootstrap_cfn/fab_tasks.py
```diff
@@ -371,3 +350,13 @@ def rsync():
...
+    tfile_path = '/tmp/{0}-temp-pillar.sls'.format(stack_name)
+    with open(tfile_path, 'w') as tfile:
+        yaml.dump(cfg, tfile) 
+        put(
+            remote_path=os.path.join(
+                remote_pillar_dir,
+                'cloudformation.sls'),
+            local_path=tfile_path,
+            use_sudo=True)
+    os.unlink(tfile_path)
```
ashb: Do we really have to create a temporary file to do this? Can we pass a StringIO object instead?

> local_path may alternately be a file-like object, such as the result of open('path') or a StringIO instance.

mattmb:
@stuartornum shouldn't need to change the example config. Good shout on the README. I need to update the CHANGELOG too.

stuartornum:
The example yaml's have no reference to salt - so I don't know (without going through the code) what the salt structure should look like

bootstrap_cfn/fab_tasks.py
mattmb: ooo, that would be nicer, will give it a try.

PR #46, opened 2015-03-23
Security Groups as dict

mattmb:
This changes the security groups section of the yaml file into a dict.
This means you can create multiple SGs with multiple rules and they can reference each other.


stuartornum:
LGTM - although I have not tested it yet... will do now

PR #48, opened 2015-03-24
salt docs

kyriakosoikonomakos:
Add some docs on salt related config

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -57,6 +57,7 @@ If you also want to bootstrap the salt master and minions, you can do this::
 
     fab application:courtfinder aws:prod environment:dev config:/path/to/courtfinder-dev.yaml cfn_create install_master install_minions
 
+
 Example Configuration
 ======================
 AWS Account Configuration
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -57,6 +57,22 @@ If you also want to bootstrap the salt master and minions, you can do this::
...
+    **Default value**: ./salt
+- **local_pillar_dir**: Directory containing all the files you want to have in your pillar root. 
+    **Default value**: ./pillar
+- **local_vendor_dir**: Directory containing formulas cloned by salt-shaker.
+    **Default value**: ./vendor
+- **remote_state_dir**: Salt root on the master.
+    **Default value**: /srv/salt
+- **remote_pillar_dir**: Pillar root on the master.
+    **Default value**: /srv/pillar
+
```
ashb, line 60: Where do you set these? Are they fab tasks? In the project YAML? (If project yaml they should probably be moved to that section)

stuartornum, line 60: what about: docs/sample-project.yaml

PR #49, opened 2015-03-24
Refactor fab_tasks file

ashb:
This was movitated by `get_config` having multiple return values and both @mattmb and I wanting to add a new (different) value - 5 is too certainly too many.

Main changes are:
- `get_config()` now returns just a ConfigParser instance
- `get_connection` helper function added to get Cloudformation, EC2, IAM etc connection objects
- PEP8 fixes.
- Removed the unused task `get_stack_addresses` (it was unused cos it didn't do what it said - it returned IDs, not IPs)


pidah:
LGTM; can we update the Changelog

PR #50, opened 2015-03-24
Enable cert updates

sevenmachines:
When we change the certificate in our cloud formation config we want a way to update the certificates on instances, and to set these certificates on any https load balancers on the stack. 
This patch creates a fab_task update_certs() to carry out both of these tasks, first removing/adding the certificates using IAM, then using a simple ELB class to set_ssl_certificates() on any https listeners on the stacks load_balancers

### DIFF
--- a/bootstrap_cfn/elb.py
+++ b/bootstrap_cfn/elb.py
@@ -1,0 +1,87 @@
+import sys
+import logging
+import boto.ec2.elb
+from bootstrap_cfn import cloudformation
+from bootstrap_cfn import iam
+from bootstrap_cfn import utils
+from bootstrap_cfn.errors import CloudResourceNotFoundError
+
+
+class ELB:
+
+    cfn = None
+    iam = None
+    aws_region_name = None
+    aws_profile_name = None
+
+    def __init__(self, aws_profile_name, aws_region_name='eu-west-1'):
+        self.aws_profile_name = aws_profile_name
+        self.aws_region_name = aws_region_name
+
+        self.conn_elb = utils.connect_to_aws(boto.ec2.elb, self)
+
+        self.iam = iam.IAM(aws_profile_name, aws_region_name)
+        self.cfn = cloudformation.Cloudformation(
+            aws_profile_name, aws_region_name
+        )
+
+    def set_ssl_certificates(self, ssl_config, stack_name):
+        """
+        Look for SSL listeners on all the load balancers connected to
+        this stack, then set update the certificate to that of the config
+
+        Args:
+            ssl_config (dictionary): Certification names to corresponding data
+            stack_name (string): Name of the stack
+
+        Returns:
+            list: The list of load balancers that were affected by the change
+
+        Raises:
+            CloudResourceNotFoundError: Raised when the load balancer key in the cloud
+                config is not found
+        """
+        updated_load_balancers = []
+        for cert_name in ssl_config.keys():
+            # Get the cert id and also its arn
+            cert_id = "{0}-{1}".format(cert_name, stack_name)
+            cert_arn = self.iam.get_arn_for_cert(cert_id)
+
+            # Get all stack load balancers
+            load_balancer_resources = self.cfn.get_stack_load_balancers(stack_name)
+            found_load_balancer_names = [lb.physical_resource_id for lb in load_balancer_resources]
+            # Use load balancer names to filter getting load balancer details
+            load_balancers = []
+            if len(found_load_balancer_names) > 0:
+                load_balancers = self.conn_elb.get_all_load_balancers(load_balancer_names=found_load_balancer_names)
+
+            # Look for https listeners on load balancers and update the cert
+            # using the arn
+            if len(load_balancers) > 0:
+                for load_balancer in load_balancers:
+                    for listener in load_balancer.listeners:
+                        # Get protocol, if https, update cert
+                        # in_port = listener[0]
+                        out_port = listener[1]
+                        protocol = listener[2]
+                        # If the protocol is HTTPS then set the cert on the listener
+                        if protocol == "HTTPS":
+                            logging.info("ELB::set_ssl_certificates: "
+                                         "Found HTTPS protocol on '%s', "
+                                         "updating SSL certificate with '%s'" 
+                                         % (load_balancer.name, cert_arn))
+                            self.conn_elb.set_lb_listener_SSL_certificate(load_balancer.name,
+                                                                          out_port,
+                                                                          cert_arn
+                                                                          )
+                            updated_load_balancers.append(load_balancer)
+
+            else:
+                # Throw key error. There being no load balancers to update is not
+                # necessarily a problem but since the caller expected there to be let
+                # it handle this situation
+                raise CloudResourceNotFoundError("ELB::set_ssl_certificates: "
+                                                 "No load balancers found in stack,")
+
+        return updated_load_balancers
+
--- a/bootstrap_cfn/fab_tasks.py
+++ b/bootstrap_cfn/fab_tasks.py
@@ -2,6 +2,8 @@
 
 import os
 import sys
+import time
+import logging
 
 from fabric.api import env, task
 from fabric.utils import abort
@@ -10,6 +12,7 @@ from fabric.colors import green, red
 from bootstrap_cfn.config import ProjectConfig, ConfigParser
 from bootstrap_cfn.cloudformation import Cloudformation
 from bootstrap_cfn.iam import IAM
+from bootstrap_cfn.elb import ELB
 from bootstrap_cfn.utils import tail
 
 
@@ -172,3 +175,42 @@ def cfn_create():
         if 'ssl' in cfn_config.data:
             iam.delete_ssl_certificate(cfn_config.ssl(), stack_name)
         abort('Failed to create stack: {0}'.format(stack))
+        
+@task
+def update_certs(force=False):
+    """
+    Update the ssl certificates with those in the config file.
+    Also handle settings the certificates on ELB's
+    """
+    if force:
+        logging.info("Forcing certificate updates is enabled...")
+
+    stack_name = get_stack_name()
+    cfn_config = get_config()
+    # Upload any SSL certificates to our EC2 instances
+    updated_count = False
+    if 'ssl' in cfn_config.data:
+        logging.info("Reloading SSL certificates...")
+        iam = get_connection(IAM)
+        updated_count = iam.update_ssl_certificates(cfn_config.ssl(),
+                                                    stack_name,
+                                                    force)
+    else:
+        logging.error("No ssl section found in cloud config file, aborting...")
+        sys.exit(1)
+
+    # Arbitrary wait to allow SSL upload to register with AWS
+    # Otherwise, we can get an ARN for the load balancer certificates
+    # without it being ready to assign
+    time.sleep(3)
+
+    # Set the certificates on ELB's if we have any
+    if updated_count > 0:
+        if 'elb' in cfn_config.data:
+            logging.info("Setting load balancer certificates...")
+            elb = get_connection(ELB)
+            updated_load_balancers = elb.set_ssl_certificates(cfn_config.ssl(), stack_name)
+    else:
+        logging.error("No certificates updated so skipping "
+                      "ELB certificate update...")
+
--- a/bootstrap_cfn/iam.py
+++ b/bootstrap_cfn/iam.py
@@ -1,8 +1,11 @@
 import boto.iam
+from boto.connection import AWSQueryConnection
 from boto.exception import NoAuthHandlerFound
 from boto.provider import ProfileNotFoundError
-
 from bootstrap_cfn import utils
+import logging
+from bootstrap_cfn.errors import CloudResourceNotFoundError
+
 
 class IAM:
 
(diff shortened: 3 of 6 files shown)


bootstrap_cfn/elb.py
```diff
@@ -0,0 +1,75 @@
...
+                            self.conn_elb.set_lb_listener_SSL_certificate(load_balancer.name,
+                                                                          out_port,
+                                                                          cert_arn
+                                                                          )
+            else:
+                # Just info, there being no load balancers to update is not
+                # necessarily a problem
+                logging.info("ELB::set_ssl_certificates: "
+                             "No load balancers found in stack, "
+                             "no certificate updates needed")
```
ashb: This probably be a warning or an error - the user has requested something to be updated that couldn't be. This also means we should likely return a non-0 exit code.

bootstrap_cfn/fab_tasks.py
```diff
@@ -317,3 +319,26 @@ def rsync():
...
+        iam.upload_ssl_certificate(cfn_config.ssl(), stack_name)
+    # Arbitrary wait to allow SSL upload to register with AWS
+    # Otherwise, we can get an ARN for the load balancer certificates
+    # without it being ready to assign
+    time.sleep(3)
+
+    # Set the certificates on ELB's
+    if 'elb' in cfn_config.data:
+        elb = ELB(aws_config)
+        elb.set_ssl_certificates(cfn_config.ssl(), stack_name)
```
ashb: Similar thing about the `if`s here - we should check the both exist to start with and error if they don't.

bootstrap_cfn/iam.py
```diff
@@ -29,7 +31,57 @@ def upload_ssl_certificate(self, ssl_config, stack_name):
...
+            cert_name (string): The name of the certification
+
+        Returns:
+            cert_arn (string): The certifications arn if found,
+                None type otherwise
+        """
+        cert_arn = None
+
+        # Collect all certificates into a list
+        certs = self.conn_iam.get_all_server_certs()
```
ashb: Can you use `iam.get_server_certificate(cert_name)` instead? http://boto.readthedocs.org/en/latest/ref/iam.html#boto.iam.connection.IAMConnection.get_server_certificate

bootstrap_cfn/iam.py
```diff
@@ -29,7 +31,57 @@ def upload_ssl_certificate(self, ssl_config, stack_name):
...
+        certlist = certs['list_server_certificates_response']['list_server_certificates_result']['server_certificate_metadata_list']
+        # Search for the certificate with the desired name in all certs
+        # and collect the arn if it exists
+        for cert in certlist:
+            if cert['server_certificate_name'] == cert_name:
+                cert_arn = cert.arn
+                break
```
ashb: When we find the result it is cleaner to just return the result directly rather than break and check again if we've found an answer. (thought this loop might go away) if we can use get_server_certificate directly.

bootstrap_cfn/fab_tasks.py
```diff
@@ -317,3 +319,26 @@ def rsync():
         remote_dir=remote_pillar_dir,
         local_dir=os.path.join(local_pillar_dir, env.environment, '.'),
         use_sudo=True)
+
+@task 
+def update_certs():
```
ashb: If possible it would be really nice if this method was idempontent - i.e. if the certs haven't changed then running this function should do nothing.

I'm not sure how feasible that is to do thought...

ashb:
Can you also rebase this into a small number of commits? In this case probably just 1.

bootstrap_cfn/elb.py
sevenmachines: I was going to throw an exception here to allow the caller to deal with there being no elbs or ignoring, exiting seems a little strict?

bootstrap_cfn/fab_tasks.py
niallcreech: If the certs dont exist then it should be a non-zero exit, but elb's aren't required to exist, so handling the exception when they dont seems ok to do.

sevenmachines: Trying to update certs when there not defined in the config warrants a non-zero exit right enough. 
The second if logic is 'Look for elb's and update them if we have them', they dont need to exist

ashb:
@niallcreech Travis is reporting that the tests are failing

```
FAIL: test_ssl_delete (tests.test.CfnTestCase)
----------------------------------------------------------------------
  File "/home/travis/build/ministryofjustice/bootstrap-cfn/tests/test.py", line 246, in test_ssl_delete
    self.assertTrue(x)
AssertionError: False is not true
```

Did you change the API here perhaps?

sevenmachines:
Yep, I'm on it. its not the API, delete checks to see if the cert exists. I'm setting up the test to mock the call to list_server_certs.

sevenmachines:
- Update test_ssl_delete test case to reflect new 'check cert exists' flow for IAM class

Now expanding tests to cover update_certs fab_task

sevenmachines:
Need to fill out the test cases then hopefully done

sevenmachines:
Added more test cases, hopefully ready for review

bootstrap_cfn/fab_tasks.py
```diff
@@ -312,3 +314,45 @@ def rsync():
...
+
+
+@task
+def update_certs(force=False, verbose=False):
+    """
+    Update the ssl certificates with those in the config file.
+    Also handle settings the certificates on ELB's
+    """
+    if verbose:
+        logging.basicConfig(level=logging.INFO)
```
mattmb: Rather than setting a flag called verbose here, it would be better to just use the logger with different levels. Like `logging.info` when you want it to appear for everyone and `logging.debug` when you want it only if the verbosity is increased. Then we can set the level globally in the fabfile maybe?

bootstrap_cfn/elb.py
```diff
@@ -0,0 +1,80 @@
...
+        for cert_name in ssl_config.keys():
+            # Get the cert id and also its arn
+            cert_id = "{0}-{1}".format(cert_name, stack_name)
+            cert_arn = self.iam.get_arn_for_cert(cert_id)
+
+            # Get all stack load balancers
+            load_balancer_resources = self.cfn.get_stack_load_balancers(stack_name)
+            found_load_balancer_names = []
+            for load_balancer_resource in load_balancer_resources:
+                found_load_balancer_names.append(load_balancer_resource.physical_resource_id)
```
mattmb: This can be made a bit more pythony like this:

```
found_load_balancer_names = [ lb.physical_resource_id for lb in load_balancer_resources]
```

bootstrap_cfn/elb.py
```diff
@@ -0,0 +1,80 @@
...
+
+            # Use load balancer names to filter gettingload balancer details
+            if len(found_load_balancer_names) > 0:
+                load_balancers = self.conn_elb.get_all_load_balancers(load_balancer_names=found_load_balancer_names)
```
mattmb, line 56: Do you know what get_all_load_balancers returns if there are none found? If it's an empty list (or iterator) then we might not need the `if len()` on line 58.

Likewise if you call it with load_balancer_names as empty it might just return an empty list and you wont need the one on 53 either.

mattmb, line 56: Note this isn't super important, just if you happen to know how it works.

mattmb, line 56: Just noticed you're using it to throw an exception so ignore me :-)

bootstrap_cfn/iam.py
```diff
@@ -18,18 +20,299 @@ def __init__(self, aws_profile_name, aws_region_name='eu-west-1'):
...
+        Returns:
+            success(bool): True if certificate is uploaded, False otherwise
+        """
+        success = False
+
+        cert_body = ssl_data['cert']
+        private_key = ssl_data['key']
+        try:
+            cert_chain = ssl_data['chain']
+        except:
```
mattmb: This should be `except KeyError` (always good to catch exceptions as specifically as possible.

bootstrap_cfn/iam.py
```diff
@@ -18,18 +20,299 @@ def __init__(self, aws_profile_name, aws_region_name='eu-west-1'):
...
+        Args:
+            cert_name(string): The name of the certificate entry to look up
+            stack_name(string): The name of the stack
+            ssl_data(dictionary): The configuration data for this certificate
+                entry
+
+        Returns:
+            success(bool): True if certificate is uploaded, False otherwise
+        """
+        success = False
```
mattmb: You can do away with this by just doing `return False` or `return True`

bootstrap_cfn/iam.py
```diff
@@ -18,18 +20,299 @@ def __init__(self, aws_profile_name, aws_region_name='eu-west-1'):
...
+            succeeded_count(int): Number of certificates that were successfully updated
+        """
+        succeeded_count = 0
+        for cert_name, ssl_data in ssl_config.items():
+            try:
+                cert_exists = self.check_remote_certificate_exists(cert_name,
+                                                                   stack_name,
+                                                                   ssl_data)
+                if cert_exists:
+                    if force:
```
mattmb: `if cert_exists and force`

bootstrap_cfn/iam.py
```diff
@@ -18,18 +20,299 @@ def __init__(self, aws_profile_name, aws_region_name='eu-west-1'):
...
+                        self.delete_certificate(cert_name,
+                                                stack_name,
+                                                ssl_data,
+                                                force)
+
+                    upload_success = self.upload_certificate(cert_name,
+                                                             stack_name,
+                                                             ssl_data,
+                                                             force)
```
mattmb: So we're doing a delete and create? Does this work when the certificate is already attached to an ELB? What happens in the small time in between when there's no cert?

mattmb: Oh I see, it's ok because the cert your deleting remains attached to the listener on the ELB?

bootstrap_cfn/iam.py
```diff
@@ -18,18 +20,299 @@ def __init__(self, aws_profile_name, aws_region_name='eu-west-1'):
...
+            cert_name(string): The name of the certificate entry to look up
+            stack_name(string): The name of the stack
+            ssl_data(dictionary): The configuration data for this
+                certificate entry
+
+        Returns:
+            exists(bool): True if remote AWS certificate exists, false otherwise
+        """
+
+        exists = False
```
mattmb: likewise with this, clearer to just `return False`

bootstrap_cfn/iam.py
```diff
@@ -18,18 +20,299 @@ def __init__(self, aws_profile_name, aws_region_name='eu-west-1'):
...
+            remote_cert_chain = remote_cert_certificate.get("certificate_chain",
+                                                            None)
+            remote_cert_key = remote_cert_certificate.get("certificate_key",
+                                                          None)
+            remote_cert_data = \
+            {
+             "cert": remote_cert_body,
+             "chain": remote_cert_chain,
+             "key": remote_cert_key,
+             }
```
mattmb: Can you make this into one block like:
`"cert": remote_cert_certificate.get("certificate_body", None),` etc

bootstrap_cfn/iam.py
```diff
@@ -18,18 +20,299 @@ def __init__(self, aws_profile_name, aws_region_name='eu-west-1'):
...
+            # Compare the local cert and chain certificates to remote
+            if self.compare_certificate_data(remote_cert_data, remote_cert_data):
```
mattmb: This looks like a mistake?

bootstrap_cfn/iam.py
```diff
@@ -18,18 +20,299 @@ def __init__(self, aws_profile_name, aws_region_name='eu-west-1'):
...
+                                                          end_text)
+            if not chains_are_equal:
+                logging.info("IAM::compare_certificate_data: "
+                             "Certificate chain data is not equal")
+            else:
+                are_equal = True
+
+        return are_equal
+
+    def compare_text_between_strings(self,
```
mattmb: Can we put the start text/end text in this method (rather than passing it in). Think you might be trying to keep it generic but in this case I feel like it would be better to make it specific to the problem. Maybe rename it to something like compare_certs_body or something like that.

mattmb:
I still need to review the tests. :smile:

bootstrap_cfn/fab_tasks.py
```diff
@@ -312,3 +314,45 @@ def rsync():
...
+                                                    stack_name,
+                                                    force)
+    else:
+        logging.error("No ssl section found in cloud config file, aborting...")
+        sys.exit(1)
+
+    # Arbitrary wait to allow SSL upload to register with AWS
+    # Otherwise, we can get an ARN for the load balancer certificates
+    # without it being ready to assign
+    time.sleep(3)
```
mattmb, line 205: Ugh. Is that a limitation of the AWS API? If so amazon suck :angry:

sevenmachines, line 205: seems to be, couldn't find another explanation for the delay in availablity

sevenmachines:
- Dropped verbose arg from fab task
- elb: Made load_balance search more pythonistic
- iam: Made exception catching more specific to KeyError
- iam: Move to pythonic mid-method returns rather than collecting
  results and returning on method exit
- iam: Combine cert_exists and force checks on if
- iam: combine some dictionary calls to compact code
- iam: Fix remote and local cert data comparison
- iam: Compact update_ssl_certificates checks on force and exists

sevenmachines:
Not sure about some of this 'forcing' logic

bootstrap_cfn/elb.py
```diff
@@ -0,0 +1,77 @@
...
+        for cert_name in ssl_config.keys():
+            # Get the cert id and also its arn
+            cert_id = "{0}-{1}".format(cert_name, stack_name)
+            cert_arn = self.iam.get_arn_for_cert(cert_id)
+
+            # Get all stack load balancers
+            load_balancer_resources = self.cfn.get_stack_load_balancers(stack_name)
+            found_load_balancer_names = [lb.physical_resource_id for lb in load_balancer_resources]
+            # Use load balancer names to filter gettingload balancer details
+            if len(found_load_balancer_names) > 0:
```
mattmb, line 55: think we might want `load_balancers=[]` here or in the else block because if this is not true the next if statement will fail (if I'm reading it right).

bootstrap_cfn/iam.py
```diff
@@ -18,18 +20,292 @@ def __init__(self, aws_profile_name, aws_region_name='eu-west-1'):
...
+            are_equal: True if the certficate data are equal,
+            false otherwise
+        """
+
+        are_equal = False
+        certs_are_equal = self.compare_certs_body(
+                                                  cert_data1.get("cert", None),
+                                                  cert_data2.get("cert", None))
+        if not certs_are_equal:
+            logging
```
mattmb: missing the log message here?

mattmb:
@niallcreech and I have just had a good chat about this.

Two things we thought of:
- Force may not be necessary. Use cases are my repo has the correct cert, I run the tool, nothing changes. My repo has a new cert, I run the tool, the cert gets updated.
- Multiple ELBs with different certs. We think it's probably safe the way it works at the moment because the cert wont be updated, the ARN wont change etc. But might be nicer to track what's changing (rather than an updated count (maybe a dictionary of what's updated)

stuartornum:
This is a massive PR, I have no idea where to start

sevenmachines:
Mostly covered with Matt yesterday, I’ve got a couple of small changes to make but unfortunately I’ve no internet until the afternoon. Press on with any bootstrap work you’re doing, I’ll fix this up around it once you’re finished

sevenmachines:
- Reworked IAM force options to remove them and have sane logic
- Return list of updated elb's when we run update_ssl's
- Updated tests to reflect new force and list return logic

sevenmachines:
Will need to wait until trim_the_fat has been merged and see how rebasey this will be before moving to review

mattmb:
Cool, thanks @niallcreech I've got my fingers crossed with the git gods that it won't be too conflicty :smile:

mattmb:
I'd say this looks good now. Anyone else want to comment before I merge?

Issue #13, opened 2015-02-24
ELB instance SG not included in EC2 instance's SG


milosgajdos:
When creating CF stack which has EC2 instance(s) and ELB, ELB SG is not included into EC2 instance's which makes EC2 unavailable from the outside via ELB.

ashb:
Fixed in #51

PR #51, opened 2015-03-24
Security Groups for ELBs

mattmb:
This adds the possibility to add security groups for your ELBs, it also defines a default ELB SG of open to the world on 80/443.


tests/tests.py
```diff
@@ -257,7 +260,7 @@ def test_elb(self):
             ProjectConfig(
                 'tests/sample-project.yaml',
                 'dev').config,'my-stack-name')
-        self.assertEquals(known, config.elb())
+        compare(known, config.elb()[0])
```
ashb: This looks like a suspect – can you assidn `elb, elb_sg` and then compare elb directly as it's clearer what this is doing.

tests/tests.py
```diff
@@ -365,7 +369,7 @@ def test_elb_with_ssl(self):
             ],
         }]
         config = ConfigParser(project_config.config,'my-stack-name')
-        self.assertEquals(known, config.elb())
+        self.assertEquals(known, config.elb()[0])
```
ashb: Same here

tests/tests.py
```diff
@@ -413,7 +418,7 @@ def test_elb_with_reserved_chars(self):
...
```
ashb: And one more.

tests/tests.py
mattmb: Good shout

ashb:
It looks like this will have a default SG on the ELB allowing 80 and 443 from everyone, but if you specify any SGs for your ELB it will be replaced and the default one won't be used anymore?

Is that right? If so can you make that clearer in the readme please?

mattmb:
Yep, you got it. Will add a comment to make it clear.

mattmb:
@ashb you're right this does allow us to give only ELBs access to the hosts, solving #13 .

See https://github.com/ministryofjustice/template-deploy/pull/5 for how.

bootstrap_cfn/config.py
```diff
@@ -225,6 +227,17 @@ def elb(self):
                     listener.update(ssl_template)
 
 
+            elb_sg = template.pop('DefaultELBSecurityGroup')
+            if 'security_groups' in elb:
+                for sg_name, sg in elb['security_groups'].items():
+                    new_sg = deepcopy(elb_sg)
+                    new_sg['Properties']['SecurityGroupIngress'] = sg
+                    elb_sgs[sg_name + safe_name] = new_sg
```
ashb: Security group names can contain "a-z, A-Z, 0-9, spaces, and ._-:/()#,@[]+=&;{}!$*" so we should be able to use the raw SG name without modification - this makes it clearer to parse the SG name - currently we would end up with something like "DefaultSGtesttemplateext" but "DefaultSG-test-template-ext" would be much preferable.

ashb: The above commend applies more to line 238 `['DefaultSG' + safe_name]`.

For this one: it's slightly unexpected that the SG name you put in the yaml isn't the SG name you'd put elsewhere to reference it...

PR #55, opened 2015-03-26
If the folder exists the current script dies

ciprianc:
This makes it do a git pull if it's already there, otherwise git clone.

Why is this a separate shell script and not in the fabric task?

### DIFF
--- a/scripts/bootstrap-salt.sh
+++ b/scripts/bootstrap-salt.sh
@@ -5,7 +5,13 @@ apt-get update
 apt-get -y install python-setuptools git
 easy_install boto
 
-cd /usr/local && git clone https://github.com/ministryofjustice/bootstrap-cfn
+# get or update bootstrap-cfn
+if [ -d "/usr/local/bootstrap-cfn" ]; then
+  cd /usr/local/bootstrap-cfn && git pull
+else
+  cd /usr/local && git clone https://github.com/ministryofjustice/bootstrap-cfn
+fi
+
 easy_install https://s3.amazonaws.com/cloudformation-examples/aws-cfn-bootstrap-latest.tar.gz
 chmod 755 /usr/local/bootstrap-cfn/scripts/ec2_tags.py
 chmod 750 /usr/local/bootstrap-cfn/bootstrap_cfn/salt_utils.py


ashb:
Reason this is a shell script not a fab task is that (I think - @mattmb or @kyriakosoikonomakos can you confirm?) it's fed into userdata and run automatically by EC2 when the instance comes up.

A random point here: This picks the version of script from master of this repo. Should we use the _same_ version of the repo that we are running from (i.e. change `git clone` to `curl https://github.com/ministryofjustice/bootstrap-cfn/archive/$}SHA_or_tag}.zip`)?

ciprianc:
In my opinion, pinning a version not only complicates things and makes them uglier, but it defeats the purpose of continuous delivery. We shouldn't be afraid of master, as whatever is there was seen by the eyes of two different people, so it should work. If it breaks stuff we quickly know and fix it. If we leave it pinned for a long time, then after many changes we will be afraid to bump the version as too much has changed. Or stuff might break and we have to go back through all commits to see which one broke things. That's maybe not the case for this script as it's pretty small, but we should avoid pinning in general. Let me know if you still want me to pin it.

ashb:
Sorry I wasn't clear - I wasn't suggesting that we should pin the version in the repo to anything specific.

More that since this is fixed at the time the autoscaling group is created my idea was to 'pin'/resolve the version when we run the cfn_create task, so that when you make changes to this script you don't have to worry about changes affecting stacks that already exist.
1. Is it clear what I'm saying?
2. if yes, what do you think?

I'm generally in favour of not pinning things but I'm trying to make it easy for us to iterate and develop without having to consider all stacks we've ever created.

ashb:
(we should continue this discussion after merging this PR - as this is unrelated to the proposed change)

kyriakosoikonomakos:
I think autoscaling has not been thought through for bootstrap_cfn in
general...

ciprianc:
Indeed the discussion is a bit outside the scope.

ashb:
Given this script is actually invoked from a fab task my point is rather moot anyway - I thought this script was fed into the userdata but it hasn't been since https://github.com/ministryofjustice/bootstrap-cfn/commit/b3353858f53867c138b3a3a493c42018881034d3

PR #64, opened 2015-04-08
 Added fix and test for missing ELB definition as per issue#56

### LINKED ISSUE #56: Regression - ELB no longer optional when it should be
I created a regression here:
https://github.com/ministryofjustice/bootstrap-cfn/blob/master/bootstrap_cfn/config.py#L84

This blows up if you have no ELB

lukaszraczylo:
Closes #56


lukaszraczylo:
Do not merge please

PR #66, opened 2015-04-09
Test custom SGs on ELBs better.

ashb:
- Add test the default behavior (i.e. elb_sgs return value)
- Have separate test case that just tests custom SGs

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -149,87 +149,72 @@ class TestConfigParser(unittest.TestCase):
         self.assertEquals(known, config.rds())
 
     def test_elb(self):
-        known = [{'ELBtestdevexternal': {'Type': 'AWS::ElasticLoadBalancing::LoadBalancer',
-                                         'Properties': {'Listeners': [{'InstancePort': 80,
-                                                                       'LoadBalancerPort': 80,
-                                                                       'Protocol': 'TCP'},
-                                                                      {'InstancePort': 443,
-                                                                       'LoadBalancerPort': 443,
-                                                                       'Protocol': 'TCP'}],
-                                                        'AvailabilityZones': {'Fn::GetAZs': ''},
-                                                        'Scheme': 'internet-facing',
-                                                        'LoadBalancerName': 'ELB-test-dev-external'}}},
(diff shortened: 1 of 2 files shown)


tests/tests.py
```diff
@@ -149,97 +149,160 @@ def test_rds(self):
...
+                    'GroupDescription': 'DefaultELBSecurityGroup'
+                },
+                'Type': u'AWS::EC2::SecurityGroup',
+            },
+        }
+
+        project_config = ProjectConfig('tests/sample-project.yaml', 'dev')
+
+        # Remove the "test-dev-internal" ELB
+        project_config.config['elb'].pop()
```
mattmb: This feels likely to break. Can we check it's name? Or pop the elb key and write your own in the test?

ashb: Good call.

tests/tests.py
```diff
@@ -149,97 +149,160 @@ def test_rds(self):
...
+                 'ToPort': 443,
+                 'CidrIp': '1.2.3.4/32'},
+            ]
+        }
+        config = ConfigParser(project_config.config, 'my-stack-name')
+        elb_cfg, elb_sgs = config.elb()
+
+        compare(expected_sgs, elb_sgs)
+
+        [elb] = (e.values()[0] for e in elb_cfg if e.values()[0]['Type'] == 'AWS::ElasticLoadBalancing::LoadBalancer')
```
mattmb: I think it might be easier to read just having the expected result and doing a compare?

mattmb:
Otherwise looks great :+1:

tests/tests.py
ashb: My thought for doing this is that: "I'm not testing the whole return value (which is about a 40 line block) but just this one specific property (And to not overly test/retest that which is already covered in `test_elb`)

Is this clearer?

``` python
        [elb] = (e.values()[0] for e in elb_cfg if e.has_key('ELBtestdevexternal'))

```

mattmb: Thought that might be what you were thinking, I'm happy. That version is more readable :smile:

ashb:
@mattmb I've aAddressed your two comments. Any other improvements?

PR #67, opened 2015-04-09
[dockerfile_for_local_execution] Intial test version

oskarpearson:
Docker file for running bootstrap in a container, so that the process of
installing dependent libraries is simplified.
- Added README on how to run things in the container


Dockerfile
```diff
@@ -0,0 +1,47 @@
+FROM ubuntu:14.04
+
+MAINTAINER WebOps MoJ <webops@digital.justice.gov.uk>
```
ashb: This should be `platforms@` as `webops@` includes people from OPG.

Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+# Install libgit2 - http://www.pygit2.org/install.html
+RUN mkdir /tmp/libgit2
+WORKDIR /tmp/libgit1
+ADD https://github.com/libgit2/libgit2/archive/v0.22.0.tar.gz /tmp/libgit1/
+RUN tar zxf v0.22.0.tar.gz
+# Perform the build
+WORKDIR /tmp/libgit1/libgit2-0.22.0
+RUN cmake .
+RUN make
+RUN make install
```
ashb: We'll need libssh2 support for salt-shaker to be able to clone most/all of our repos

Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+
+# Copy and work on just the requirements.txt file, so that the 'pip install' step is cacheable
+COPY requirements.txt /tmp/requirements.txt
+RUN pip install -r /tmp/requirements.txt
+
+# Cleanup
+RUN rm -rf /tmp/libgit2
+RUN rm -rf /var/lib/apt/lists/*
+RUN rm -fr *Release* *Sources* *Packages*
+RUN truncate -s 0 /var/log/*log
```
ashb: Unless things have changed recently this cleanup step doesn't actually make the resulting docker image any smaller as these files are cached in the upper layers :(

 Thanks Docker. Thocker.

oskarpearson: Very good point. Hmm. I based this on https://github.com/ministryofjustice/docker-templates/blob/48de00e046bedf6893c51aca7fab82f6e00d4a1a/socat-ssh/Dockerfile.

I'm guessing then that if we chain those actions onto the end of the apt-get install, then the layer will be cleaned before it's committed.

I'm going to commit a change for this shortly

ashb: Yup - making them all part of one `RUN` stanza means docker only commits the FS layer after running all of them.

oskarpearson:
Currently this doesn't support git over SSH - so it's pending. I'm currently focussed elsewhere, but will come back to it as soon as I can.

This needs to work in the container:

```
python -c 'import pygit2; print pygit2.features & pygit2.GIT_FEATURE_SSH'
```

ashb:
Remind me to tell you about our deb pkg repo - that should have a libgit2 (maybe called dsd-libgit2) that is compiled with libssh2 support

PR #70, opened 2015-04-10
show cloudformation events while creating and deleting the stack

ciprianc:


pidah:
@ciprianc works nicely; can we move get_events and tail functions from fab_tasks into a separate module ?
And add some tests.

ashb:
I think we said that we were happy with this without testing but we just thought it should be in a separate file (was that so the only functions in fab_tasks.py are actual fab tasks?)

stuartornum:
@ashb - yep, thats correct

pidah:
LGTM :+1:

pidah:
reopen

PR #71, opened 2015-04-10
IAM cross account

mattmb:
In order for a central platform to build/deploy to stacks created with this tool we need some AWS API access so that we can:
- Find the master's IP address based on the tag created
- Upload static assets to the S3 bucket (if created)

This PR adds some **optional** IAM roles that can be assumed from another account. This makes it possible for a deployment script to get the access described above from a limited user in a central AWS account.


bootstrap_cfn/config.py
```diff
@@ -57,6 +57,7 @@ def process(self):
         elb = {}
         output_templates = []
 
+        output_templates.append('stacks/iam_out.json')
```
mattmb: Just noticed this needs an if block...

bootstrap_cfn/stacks/iam-deploy.json
```diff
@@ -0,0 +1,74 @@
+{
+    "CrossAccountRole" : {
+        "Type" : "AWS::IAM::Role",
+        "Properties" : {
+            "AssumeRolePolicyDocument" : {
+                "Statement" : [
+                     {
+                         "Effect" : "Allow",
+                         "Principal" : {
+                             "AWS": "arn:aws:iam::ACCOUNT_NUMBER_HERE:root"
```
ashb, line 10: This looks suspect

ashb, line 10: Oh I see, this is munged by the config loader.

Rather than doing this and then changing the template inline how about we refer to a CFN Parameter. In the base.json:

``` json
"Parameters" : {
  "DeployAccountID" : {
    "Type" : "Number",
    "Description" : "Put a useful desc here"
  }
}
```

Then in the stack this bit becomes:

``` json
                         "Principal" : {
                             "AWS": { "Fn::Join" : ["",
                                 "arn:aws:iam::",
                                 { "Ref": "DeployAccountID" }
                                 ":root"
                             ] }
                         },
```

I think this makes it clearer - (it means that our python code has to do less at least)

bootstrap_cfn/config.py
```diff
@@ -99,7 +102,22 @@ def vpc(self):
...
+                deploy_iam = json.loads(pkgutil.get_data('bootstrap_cfn', 'stacks/iam-deploy.json'))
+                statement = deploy_iam['CrossAccountRole']['Properties']['AssumeRolePolicyDocument']['Statement']
+                statement[0]['Principal']['AWS'] = "arn:aws:iam::{0}:root".format(account_id)
+                if not 's3' in self.data:
+                    deploy_iam.pop('S3DeployUserPolicy')
+                else:
+                    bucket_name = self.data['s3']['static-bucket-name']
+                    policy = deploy_iam['S3DeployUserPolicy']['Properties']['PolicyDocument']['Statement']
+                    statement = policy[-1]
+                    statement['Resource'] = ["arn:aws:s3:::{0}/*".format(bucket_name), "arn:aws:s3:::{0}".format(bucket_name)]
```
ashb, line 118: Is there any way we can do more of this in the JSON templates too?

bootstrap_cfn/stacks/iam-deploy.json
ashb, line 10: This is open for discussion if anyone disagrees

mattmb, line 10: Hmm, that would make the code cleaner. To be clear you'd update the paramaters section from the code and just Ref it in the IAM section?

bootstrap_cfn/config.py
mattmb, line 118: Now that you mention it, surely we can Ref the name of the S3 bucket too...

Will take a look :+1:

bootstrap_cfn/stacks/iam-deploy.json
ashb, line 10: I think you just specify the parameters in the API call to CreateStack. Checking

ashb, line 10: Yup http://boto.readthedocs.org/en/latest/ref/cloudformation.html#boto.cloudformation.connection.CloudFormationConnection.create_stack 

> Parameters:   
> -  **parameters** (list) – A list of key/value tuples that specify input parameters for the stack.

stuartornum:
Spoke with @mattmb about this - think we may try to solve another way, leaving bootstrap-cfn purely for AWS provisioning

ashb:
Setting up the env to be able to deploy to it is part of provisioning to me, but if you can come up with a way that you are feel is better then sounds good :racehorse:

stuartornum:
@ashb - bootstrap-cfn is about provisioning of AWS services, template-deploy is about deploy.

mattmb:
We were thinking it might be nicer to setup includes. So that we can specify a json file that gets merged with the CF json. That way stuff like this could just be flat file in the template-deploy repo.

Only thing I need to check is that Ref'ing the S3 bucket name is possible.

mattmb:
Closing this and tackling the issue with #74 instead.

PR #72, opened 2015-04-13
Making block_devices non mandatory in EC2 block definition.

### LINKED ISSUE #3: block_devices should not be mandatory in ec2 section
If I don't care about attached volume to an instance I'm trying to create I should not need `block_devices` section/parameter to be present in my configuration file. Some default value should be used:

Example (incomplete) config snippet:

``` yaml
dev:
  ec2:
    auto_scaling:
      desired: 1
      max: 3
      min: 0
    tags:
      Project: docker-registry
      Role: docker
      App: docker-registry
      Env: dev
    parameters:
      KeyName: default
      InstanceType: t2.medium
```

Error:

``` bash
  File "/Users/milosgajdos/projects/bootstrap-cfn/lib/python2.7/site-packages/fabric/main.py", line 743, in main
    *args, **kwargs
  File "/Users/milosgajdos/projects/bootstrap-cfn/lib/python2.7/site-packages/fabric/tasks.py", line 424, in execute
    results['<local-only>'] = task.run(*args, **new_kwargs)
  File "/Users/milosgajdos/projects/bootstrap-cfn/lib/python2.7/site-packages/fabric/tasks.py", line 174, in run
    return self.wrapped(*args, **kwargs)
  File "/Users/milosgajdos/repos/bootstrap-cfn/fabfile.py", line 78, in cfn_create
    stack = cfn.create(stack_name, cfn_config.process())
  File "/Users/milosgajdos/repos/bootstrap-cfn/helpers/config.py", line 49, in process
    ec2 = self.ec2()
  File "/Users/milosgajdos/repos/bootstrap-cfn/helpers/config.py", line 189, in ec2
    for i in self.data['ec2']['block_devices']:
KeyError: 'block_devices'
```

Something like 10G seems reasonable enough to me.
## Acceptance
- [ ] Sane defaults provided for block_devices secton
- [ ] Tests added to cover this case
- [ ] Docs added to readme.

lukaszraczylo:
This fixes #3

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -545,10 +545,11 @@ class TestConfigParser(unittest.TestCase):
     def test_ec2_with_no_block_device_specified(self):
         from testfixtures import compare
         project_config = ProjectConfig('tests/sample-project.yaml', 'dev')
-        project_config.config.pop('rds')
         project_config.config['ec2'].pop('block_devices')
         config = ConfigParser(project_config.config, 'my-stack-name')
-        config.process()
+        config_output = config.ec2()['BaseHostLaunchConfig']['Properties']['BlockDeviceMappings']
+        known = [{'DeviceName': '/dev/sda1', 'Ebs': {'VolumeSize': 20}}]
+        self.assertEquals(known, config_output)
 
 if __name__ == '__main__':
     unittest.main()
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -542,6 +542,13 @@ class TestConfigParser(unittest.TestCase):
                 'dev').config, 'my-stack-name')
         compare(known, config.ec2())
 
+    def test_ec2_with_no_block_device_specified(self):
+        from testfixtures import compare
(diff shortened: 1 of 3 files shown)


tests/tests.py
```diff
@@ -542,6 +542,13 @@ def test_ec2(self):
                 'dev').config, 'my-stack-name')
         compare(known, config.ec2())
 
+    def test_ec2_with_no_block_device_specified(self):
+        from testfixtures import compare
+        project_config = ProjectConfig('tests/sample-project.yaml', 'dev')
+        project_config.config.pop('rds')
+        project_config.config['ec2'].pop('block_devices')
+        config = ConfigParser(project_config.config, 'my-stack-name')
+        config.process()
```
ashb: In this case the behavior is entirely in the `ec2()` function so could you change the test to 1) call that, and 2) check the return value (or at least the devices section) is defaulted appropriately.

PR #73, opened 2015-04-13
Make the VPC configurable from with the YAML file

stuartornum:
If no VPC is specified in the YAML config, it will default to
10.0.0.0/16.

Also updated the sample-project.yaml to include this change, and instead
of copy/pasting the sample yaml to the README, we now link to it
directly.


mattmb:
LGTM :ship: :ship: :ship:

PR #74, opened 2015-04-13
Cross account IAM deployment

mattmb:
This PR adds several features:
- It adds the ability to authenticate using cross account AWS IAM roles.
- It adds the ability to include custom cloudformation templates.
- It makes the deploy script runnable as a limited user.

The overall goal is that we can create an IAM user in a central AWS account which will get limited permissions by way of an included template. The user will then be able to login using the new cross account authentication. Finally the user will be able to run the salt_utils script as a limited sudoer.

This is we think cleaner than #71 and allows us to move the IAM roles into template deploy with https://github.com/ministryofjustice/template-deploy/pull/23


mattmb:
@ashb @stuartornum here's the new approach as we discussed for review.

README.rst
```diff
@@ -109,3 +109,12 @@ By default the ELBs will have a security group opening them to the world on 80 a
...
 The certificate will be uploaded before the stack is created and removed after it is deleted.
+
+Includes
+++++++++++
+If you wish to include some static cloudformation json and have it merged with the template generated by bootstrap-cfn. You can do the following in your template yaml file::
+
+    includes:
+      - /path/to/cloudformation.json
+
+The tool will then merge this with the generated template *overwriting* any keys in the original template that clash.
```
ashb: This isn't quite clear on two counts:

1) What's a key? (Are we talking a matching logical resource id)
2) That we do deep merging (i.e. we're not overwriting the Output section entirely.)

bootstrap_cfn/salt_utils.py
```diff
@@ -1,4 +1,6 @@
 #!/usr/bin/env python
+import sys
+sys.path.append('/usr/local/bootstrap-cfn')
```
ashb: Hard coded paths are bad - it makes assumptions about where the code will live in multiple places in the code. I think this `os.path.dirname(__file__)` achieves what we want.

Also can you add a comment saying _why_ we are doing this so we remember in 12 months time.

bootstrap_cfn/config.py
```diff
@@ -92,6 +76,10 @@ def process(self):
         template['Outputs'] = {}
         for t in output_templates:
             template['Outputs'].update(json.loads(pkgutil.get_data('bootstrap_cfn', t)))
+        if 'includes' in self.data:
+            for inc_path in self.data['includes']:
+                inc = json.load(open(inc_path))
+                utils.dict_merge(template, inc)
```
ashb: You aren't doing anything with the return value from `dict_merge` - which means that it's merging in place into template - if that's the case can you remove the return statement from that function and add a simple doc comment.

ashb: Hmmm that wouldn't work actually - as it's recursive and relies on the return value inside itself. Hmmm!

ashb: Just do `template = utilts.dict_merge(template, inc)` I guess.

ashb:
The few minor comments above not withstanding this PR actually does three completely separate things
1. It adds the ability to include static snippets of Cloudformation JSON
2. It lets you use the sts AssumeRole to target another account
3. It fixes the PYTHONPATH issue in the deploy script.

These happen to be related since you want them all, but even your commits show that it's three features (the first three commits are point 1, next is point 2 and last is point 3). Sorry to be a pain but can you split into three different PRs please?

ashb:
(Or at least come up with a better PR title encompasses all the features)

README.rst
mattmb: Will try to make it clearer. Quite hard to explain deep merging in English though!

bootstrap_cfn/salt_utils.py
mattmb: Good shout, actually needs `sys.path.append(os.path.dirname(__file__) + '/..')` I think the folder of the module is always in sys.path.

bootstrap_cfn/config.py
mattmb: Ugh, I don't want to think to hard about why this seems to work because thinking about recursive functions always gives me a headache! I think there might be an edge case where it doesn't work if you don't assign the return value though so I will change it.

mattmb:
Yeah, wasn't sure whether to split or not. I feel like it's kind of one story so nice to have in one place but they are quite distinct features. I think there's some good history in this PR, so I'd rather leave it be if you're happy for me to rename it and tidy up the description :-)

ashb:
Title is much better now :+1:.

It looks like you need to rebase this onto master at least - possibly squash a few of the commits too in the process (but not all of them)

Also can you document AWS_ROLE_ARN_ID (and possibly how you would get it) and add a changelog entry for this feature.

mattmb:
@ashb rebased and tidy up :smile:

Issue #54, opened 2015-03-26
Document the custom S3 bucket policy feature


ashb:
You can specify a `policy` under the s3 bucket to override the default policy - this needs documenting.

PR #75, opened 2015-04-14
Fix/custom s3 policy

pidah:
This bugfix ensures you can apply a custom s3 policy to the cloudformation yaml config.

Connects to #54


stuartornum:
Does this work for multiple buckets? I like it generally.

However, we don't want people defining S3 storage for things that are critical to keep, for example Tribunal Decisions documents, we don't want to be declaring this within the stack. Purely because if we delete the stack, we delete the documents...

ashb:
This is just a bug fix on the existing feature that was already there, not a new feature.

It's worth noting that Cloudformation won't let you delete a stack if there are objects in the S3 bucket

stuartornum:
What I mean is the creation of the S3 bucket for Tribunal decision documents should still be a manual task - therefore not controlled by Cloudformation, as this'll be very dangerous

stuartornum:
@ashb - thats true about deleting S3 buckets with objects - forgot about that!

Issue #80, opened 2015-04-20
Support RDS encryption


mattmb:
https://github.com/ministryofjustice/bootstrap-cfn/blob/master/bootstrap_cfn/config.py#L127

At the moment we can't enable RDS encryption (StorageEncrypted) because it's not in the required fields list (see above).

Either we need to PR it in and then add it to **all** the yaml files. Or we need to support merging in non-required fields.

stuartornum:
Add it to required fields I think - makes the options clearer

mattmb:
And then every project has to update their yaml file?

PR #81, opened 2015-04-20
 Add RDS storage encryption:

pidah:
AWS Cloudformation API now supports RDS storage encryption. This
change adds StorageEncrypted key to boootstrap_cfn RDS config as
a required field.

Connected to #80

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -108,3 +108,15 @@ If you wish to include some static cloudformation json and have it merged with t
       - /path/to/cloudformation.json
 
 The tool will then perform a deep merge of the includes with the generated template dictionary. Any keys or subkeys in the template dictionary that clash will have their values **overwritten** by the included dictionary or recursively merged if the value is itself a dictionary.
+
+
+Enabling RDS encryption
++++++++++++++++++++++++++
+You can enable encryption for your DB by adding the following::
+ 
+  rds:
+     storage-encrypted: true
+     instance-class: db.m3.medium
+
+**NOTE:** AWS does not support RDS encryption for the *db.t2.** instance classes. More details on supported instance classes are available `here <http://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Overview.Encryption.html>`_
+
--- a/bootstrap_cfn/config.py
+++ b/bootstrap_cfn/config.py
@@ -159,6 +159,10 @@ class ConfigParser:
             'multi-az': 'MultiAZ'
         }
 
+        optional_fields = {
+            'storage-encrypted': 'StorageEncrypted',
+        }
+
         # LOAD STACK TEMPLATE
         template = json.loads(
             pkgutil.get_data(
@@ -173,6 +177,11 @@ class ConfigParser:
             else:
                 template['RDSInstance']['Properties'][
                     required_fields[i]] = self.data['rds'][i]
+
+        for i in optional_fields.keys():
+            if i in self.data['rds'].keys():
+                template['RDSInstance']['Properties'][
+                    optional_fields[i]] = self.data['rds'][i]
 
         return template
 
--- a/bootstrap_cfn/stacks/rds.json
+++ b/bootstrap_cfn/stacks/rds.json
@@ -37,6 +37,7 @@
       "MasterUsername": "",
       "MasterUserPassword": "",
       "AllocatedStorage": "",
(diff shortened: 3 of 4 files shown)


mattmb:
Let's add a dictionary called `optional_fields`

Then the code should not blow up if it's missing but add it if it does exist in the config.

This way we keep the nice formatting of variable names and don't break the backwards compatibility.

The only downside is that we have to PR in any new fields that we want to add.

pidah:
I think the change minimally breaks backward compatibility, with a minimal code change. Perhaps we can accomodate this breaking change ?

stuartornum:
LGTM - any reservations if I merge?

mattmb:
At the least it needs an UPGRADING.md to let people know that it's a non backward compatible change.

But I'd suggest we do what I explained above. Shouldn't be a big change.

ashb:
I think this change makes it optional. Given it is so simple lets maintain backwards compat please:

``` diff
diff --git a/bootstrap_cfn/config.py b/bootstrap_cfn/config.py
index 2a0b55b..9cd12d5 100644
--- a/bootstrap_cfn/config.py
+++ b/bootstrap_cfn/config.py
@@ -128,7 +128,6 @@ class ConfigParser:
             'db-name': 'DBName',
             'storage': 'AllocatedStorage',
             'storage-type': 'StorageType',
-            'storage-encrypted': 'StorageEncrypted',
             'backup-retention-period': 'BackupRetentionPeriod',
             'db-master-username': 'MasterUsername',
             'db-master-password': 'MasterUserPassword',
@@ -138,6 +137,9 @@ class ConfigParser:
             'instance-class': 'DBInstanceClass',
             'multi-az': 'MultiAZ'
         }
+        optional_fields = {
+            'storage-encrypted': 'StorageEncrypted',
+        }

         # LOAD STACK TEMPLATE
         template = json.loads(pkgutil.get_data('bootstrap_cfn', 'stacks/rds.json'))
@@ -151,6 +153,11 @@ class ConfigParser:
                 template['RDSInstance']['Properties'][
                     required_fields[i]] = self.data['rds'][i]

+        for i in optional_fields.keys():
+            if i in self.data['rds'].keys():
+                template['RDSInstance']['Properties'][
+                    required_fields[i]] = self.data['rds'][i]
+
         return template

     def elb(self):
diff --git a/bootstrap_cfn/stacks/rds.json b/bootstrap_cfn/stacks/rds.json
index d3ef2d7..6fb615b 100644
--- a/bootstrap_cfn/stacks/rds.json
+++ b/bootstrap_cfn/stacks/rds.json
@@ -29,7 +29,7 @@
       "MasterUserPassword": "",
       "AllocatedStorage": "",
       "StorageType": "",
-      "StorageEncrypted": "",
+      "StorageEncrypted": false,
       "Engine": "",
       "EngineVersion": "",
       "MultiAZ": "",
```

ashb:
Peter and I are looking at this... something might not be quite right yet :(

README.rst
```diff
@@ -108,3 +108,14 @@ If you wish to include some static cloudformation json and have it merged with t
...
+
+Enabling RDS encryption
++++++++++++++++++++++++++
+You can enable encryption for your DB by adding the following::
+ 
+  rds:
+     storage-encrypted: true
+     instance-class: db.m3.medium
+
+**NOTE:** AWS does not support RDS encryption for the *db.t2.** instance classes. More details on supported instance classes are available [here](http://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Overview.Encryption.html).
```
ashb: This is markdown link syntax but in an RST file.

(Stupid RST).

README.rst
```diff
@@ -108,3 +108,15 @@ If you wish to include some static cloudformation json and have it merged with t
...
+
+Enabling RDS encryption
++++++++++++++++++++++++++
+You can enable encryption for your DB by adding the following::
+ 
+  rds:
+     storage-encrypted: true
+     instance-class: db.m3.medium
+
+**NOTE:** AWS does not support RDS encryption for the *db.t2.** instance classes. More details on supported instance classes are available `here <http://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Overview.Encryption.html/>`_
```
ashb: Trailing slash on the url makes this redirect to the wrong place.

PR #84, opened 2015-04-21
Initial fat trimming of Salty'ness

stuartornum:
Tested with with the individual bootstrap-cfn CI job and then tested with template-deploy using the template yaml files


mattmb:
Can we also go through the README?

PR #86, opened 2015-04-24
DNSNAME instead of CanonicalHostedZoneName:

pidah:
When _internal_ is specified for the ELB scheme, the
load balancer doesn't have a **CanonicalHostedZoneName** value.
**DNSName** value exists for both _internal_ and _internet-facing_
ELB schemes.

### DIFF
--- a/tests/sample-project.yaml
+++ b/tests/sample-project.yaml
@@ -45,7 +45,7 @@ dev:
           Protocol: TCP
     - name: test-dev-internal
       hosted_zone: kyrtest.pf.dsd.io.
-      scheme: internet-facing
+      scheme: internal
       listeners:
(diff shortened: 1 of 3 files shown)


tests/sample-project.yaml
```diff
@@ -45,7 +45,7 @@ dev:
           Protocol: TCP
     - name: test-dev-internal
       hosted_zone: kyrtest.pf.dsd.io.
-      scheme: internet-facing
+      scheme: internal
```
stuartornum, line 48: I tried to get this to work before (internal), but for some reason it wouldn't - does it now?

stuartornum, line 48: Just read the commit notes: Awesome!

PR #89, opened 2015-04-27
Storage-type bug fix within VPC

### LINKED ISSUE #88: RDS `storage-type` ignored due to using `DBSecurityGroups`
This makes total sense AWS ?!?! (Not)

http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-rds-database-instance.html#cfn-rds-dbinstance-storagetype

For some random reason if you use both DBSecurityGroups and storage-type in the stack YAML (which we do in bootstrap-cfn), it results in storage-type being ignored and it will take the default (magnetic) instead.

stuartornum:
This now requires the stack YAML to have a "DatabaseSG" security group where you can define the port and ingress rules for the RDS instance - tested using Jenkins CI and the unit tests are passing too.

Fixes #88


ashb:
You're added a new resource called `DatabaseSG` that you refer to but I can't see it created anywhere? Am I missing something

docs/sample-project.yaml
```diff
@@ -35,6 +35,11 @@ dev:
           FromPort: 4505
           ToPort: 4506
           CidrIp: 10.0.0.0/16
+      DatabaseSG:
```
stuartornum: @ashb  - here

ashb: Isn't this going to create the SG on everything in the auto scaling group where as before it was created and used by the RDS.

ashb:
More generally I don't see the need to change the format of the config, require something new, and break backwards compatibility when all we need to do is something like this:

``` diff
@@ -33,7 +20,7 @@
       "EngineVersion": "",
       "MultiAZ": "",
       "DBSubnetGroupName": {"Ref" : "RDSSubnetGroup"},
-      "DBSecurityGroups": [{"Ref" : "DBSecurityGroup"}],
+      "VPCSecurityGroups" : [ { "Fn::GetAtt": [ "DBSecurityGroup", "GroupId" ] } ],
       "PubliclyAccessible": false,
       "AllowMajorVersionUpgrade": false,
       "AutoMinorVersionUpgrade": false
```

(i.e. use the same SG we created before, just refer to it in the right way).

stuartornum:
@ashb - thats a DBSecurityGroup, these don't work in VPC's - it needs to be an EC2 Security Group - hence the whole change

PR #91, opened 2015-04-28
Don't override fabric env vars if they are already set.

ashb:
If you invoke fabric with `fab --set aws=foo my_task` the previous way
we were setting these would then _overwrite_ the values you'd set on the
command line with lots of None and leading to hard to diagnose
behaviour.

(It also sets these at import time which could be an extra source of
confusion.)

### DIFF
--- a/bootstrap_cfn/fab_tasks.py
+++ b/bootstrap_cfn/fab_tasks.py
@@ -13,12 +13,16 @@ from bootstrap_cfn.iam import IAM
 from bootstrap_cfn.utils import tail
 
 
+# Default fab config. Set via the tasks below or --set
+env.setdefault('application')
+env.setdefault('environment')
+env.setdefault('aws')
+env.setdefault('config')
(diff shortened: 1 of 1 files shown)


bootstrap_cfn/fab_tasks.py
```diff
@@ -13,12 +13,15 @@
 from bootstrap_cfn.utils import tail
 
 
+# Default fab config. Set via the tasks below or --set
+env.setdefault('application')
+env.setdefault('environment')
+env.setdefault('aws')
+env.setdefault('config')
+env.setdefault('password')
```
ashb, line 21: This should be `stack_passwords` - the env.password was wrong before.

pidah:
env.setdefault('password') should be env.setdefault('passwords')

ashb:
@pidah I think it should be `stack_passwords` as that is what the password task writes to?

pidah:
LGTM

PR #93, opened 2015-04-30
 bump to version 0.3.3

pidah:


ashb:
Are there any other changes in flight that we should hold off for?

PR #97, opened 2015-05-05
[enable_cert_updates]

sevenmachines:
Fix for a little problem introduced in enable_cert_updates merge
- Remove deprecated force argument being passed to
  iam::delete_certificate

### DIFF
--- a/bootstrap_cfn/fab_tasks.py
+++ b/bootstrap_cfn/fab_tasks.py
@@ -105,6 +105,7 @@ def get_config():
     cfn_config = ConfigParser(project_config.config, get_stack_name())
     return cfn_config
 
+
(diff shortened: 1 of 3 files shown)


mattmb:
Cool, thanks. Can you do a `git grep force` or equivalent (just to check we didn't miss any more)

sevenmachines:
Made a few little removals to get rid of redundant 'force' entries and avoid confusion

ashb:
To test this flake8 behaviour locally make sure you have latest version installed: `pip install -U flake8`

Then you can test it like this:

```
git diff origin/master...origin/fix_deprecated_delete_certificate_force | flake8 --diff
```

PR #100, opened 2015-05-08
check that config file exists in fab env validation [issue 98]

### LINKED ISSUE #98: Handle config file not found more gracefully
If you run `fab config:i-dont-exist.yml` you get this behaviour:

```
  File "/Users/aberlin/.virtualenvs/cbo-deploy/lib/python2.7/site-packages/fabric/main.py", line 743, in main
    *args, **kwargs
  File "/Users/aberlin/.virtualenvs/cbo-deploy/lib/python2.7/site-packages/fabric/tasks.py", line 424, in execute
    results['<local-only>'] = task.run(*args, **new_kwargs)
  File "/Users/aberlin/.virtualenvs/cbo-deploy/lib/python2.7/site-packages/fabric/tasks.py", line 174, in run
    return self.wrapped(*args, **kwargs)
  File "/Users/aberlin/.virtualenvs/cbo-deploy/lib/python2.7/site-packages/bootstrap_cfn/fab_tasks.py", line 119, in cfn_delete
    cfn_config = get_config()
  File "/Users/aberlin/.virtualenvs/cbo-deploy/lib/python2.7/site-packages/bootstrap_cfn/fab_tasks.py", line 102, in get_config
    passwords=env.stack_passwords)
  File "/Users/aberlin/.virtualenvs/cbo-deploy/lib/python2.7/site-packages/bootstrap_cfn/config.py", line 16, in __init__
    self.config = self.load_yaml(config)[environment]
TypeError: 'NoneType' object has no attribute '__getitem__'
```

This is _far_ from optimal and we should check for the file not existing instead.

bennythejudge:
Fixes #98

if the config file doesn't exist, fab dies with a error message:

```
11:01 $ fab config:this-file-does-not-exist.yaml application:crimebillingonline01 aws:cbo environment:dev  passwords:./cloudformation/crimebillingonline-secrets.yaml  cfn_create  -i ~/.ssh/ubuntu.pem -u ubuntu

[ERROR] Config file this-file-does-not-exist.yaml does not exist
```

### DIFF
--- a/bootstrap_cfn/fab_tasks.py
+++ b/bootstrap_cfn/fab_tasks.py
@@ -78,17 +78,15 @@ def get_stack_name():
 
 def _validate_fabric_env():
     if env.aws is None:
-        print "\n[ERROR] Please specify an AWS account, e.g 'aws:dev'"
-        sys.exit(1)
+        sys.exit("\n[ERROR] Please specify an AWS account, e.g 'aws:dev'")
     if env.environment is None:
-        print "\n[ERROR] Please specify an environment, e.g 'environment:dev'"
-        sys.exit(1)
+        sys.exit("\n[ERROR] Please specify an environment, e.g 'environment:dev'")
     if env.application is None:
-        print "\n[ERROR] Please specify an application, e.g 'application:peoplefinder'"
-        sys.exit(1)
+        sys.exit("\n[ERROR] Please specify an application, e.g 'application:peoplefinder'")
     if env.config is None:
-        print "\n[ERROR] Please specify a config file, e.g 'config:/tmp/sample-application.yaml'"
-        sys.exit(1)
+        sys.exit("\n[ERROR] Please specify a config file, e.g 'config:/tmp/sample-application.yaml'")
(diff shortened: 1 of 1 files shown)


bootstrap_cfn/fab_tasks.py
```diff
@@ -89,6 +89,8 @@ def _validate_fabric_env():
     if env.config is None:
         print "\n[ERROR] Please specify a config file, e.g 'config:/tmp/sample-application.yaml'"
         sys.exit(1)
+    elif not os.path.isfile(env.config):
+        sys.exit("\n[ERROR] Config file %s does not exist" % str(env.config))
```
ashb, line 89: This doesn't follow the same pattern as the others - if you think this pattern is better (and it looks like it) then can you introduce another commit to change all the other ones in this function to follow the same pattern.

bennythejudge:
sys.exit can also accept a string, in which case it prints the string to STDERR and exits with 1
Within the function I have replaced the 2 lines to print and exit with just one line
Source: https://docs.python.org/2/library/sys.html#sys.exit

Issue #103, opened 2015-05-21
Convert to Troposphere to build template


ashb:
We can replace all the manual munging of JSON and referencing structures three levels deep with Troposphere.

Ash made a start by converting each of the `elb`, `rds`, etc. methods use Troposphere objects to build things but those still return dicts so we can tidy it up and remove some of the hacks – i.e. we could have each of those functions take a troposphere Template object and add the resources/outputs/mappings directly.
## Acceptance
- [x] No JSON stacks left in the project
- [x] Everything else should remain backwards compatible, no config changes or interface changes
- [x] Internals of bootstrap-cfn use Troposphere properly and don't return plain python dict objects

ashb:
Last part competed by #114

PR #104, opened 2015-05-22
Use Troposphere to generate Cloudformation, rather than munge JSON

ashb:
Replace the JSON stacks with Troposphere objects

We can and should tidy up the elb, s3, iam etc. functions to operate on a Troposphere template object directly rather than returning JSON blobs, but doing this as I have for now makes the change smaller and hopefully easier to review

Connects to  #103 but there is probably more work there to be done


ashb:
I need to test this with CFN. I have run the unit tests but our test covage is not the greatest.

ashb:
We're good - it creates a stack http://ci.dsd.io/view/Platforms/job/template-deploy_integration/61/console (that job then goes on to fail but that's because I didn't install salt-shaker properly in that virtual environment)

ashb:
I've tested this with CBO and it built and came up fine with no changes to the deploy repo. 

A diff between the stacks is here https://gist.github.com/ashb/faab9919795eddbe6bfb

The major difference is the new stack has more features, and some params are set as `"false"` instead of `false` - they still work though.

mattmb:
troposphere in requirements.txt

bootstrap_cfn/config.py
```diff
@@ -91,13 +97,170 @@ def process(self):
...
+        from troposphere import Template
+
+        t = Template()
+
+        t.add_mapping("AWSRegion2AMI", {
+            "eu-west-1": {"AMI": "ami-f0b11187"},
+        })
+
+        if 'vpc' in self.data:
+            t.app_mapping("SubnetConfig", {
```
mattmb: `add_mapping` ?

mattmb:
I like this. I've just done a diff with some of the fee remission stacks and found basically no differences, just the `false` `"false"` thing and a few description fields.

Will go through the code too now.

bootstrap_cfn/config.py
```diff
@@ -91,13 +97,170 @@ def process(self):
...
+
+        subnet_c_route_assoc = SubnetRouteTableAssociation(
+            "SubnetRouteTableAssociationC",
+            SubnetId=Ref(subnet_c),
+            RouteTableId=Ref(route_table),
+        )
+
+        resources = [vpc, subnet_a, subnet_b, subnet_c, igw, gw_attachment,
+                     public_route, route_table, subnet_a_route_assoc,
+                     subnet_b_route_assoc, subnet_c_route_assoc]
```
mattmb, line 221: This feels a little bit ugly? Would `resources.append(SubnetRouteTableAssociation(...` be better?

bootstrap_cfn/config.py
```diff
@@ -91,13 +97,170 @@ def process(self):
...
+
+        # Hack until we return troposphere objects directly
+        return json.loads(json.dumps(dict((r.title, r) for r in resources), cls=awsencode))
```
mattmb, line 224: Is this mostly to keep the tests happy? Or is it actually hard to merge together template objects?

bootstrap_cfn/config.py
```diff
@@ -230,109 +441,161 @@ def elb(self):
...
+            KeyName=self.data['ec2']['parameters']['KeyName'],
+            SecurityGroups=[Ref(g) for g in sgs],
+            InstanceType=self.data['ec2']['parameters']['InstanceType'],
+            AssociatePublicIpAddress=True,
+            IamInstanceProfile=Ref("InstanceProfile"),
+            ImageId=FindInMap("AWSRegion2AMI", Ref("AWS::Region"), "AMI"),
+            BlockDeviceMappings=devices,
+            UserData=Base64(Join("", [
+                "#!/bin/bash -xe\n",
+                "#do nothing for now",
```
mattmb, line 583: Might be nice to split this into a separate file (not a deal breaker)? I know we're not using it right now, but just so that it's not difficult later.

mattmb:
Otherwise LGTM :+1: :smile:

bootstrap_cfn/config.py
ashb, line 224: Yeah, this is to keep the function implementation exactly the same and to minimize the changes as much as possible in this PR - once we've merged this then we can change each of these functions to either return troposphere resources directly or take a Template object and add resources.

bootstrap_cfn/config.py
ashb, line 221: `[].append` returns `None` so since I'm referring to that object later in this method I can't use `.append` directly.

mattmb, line 221: Ah, gotcha. Fair enough then.

bootstrap_cfn/config.py
mattmb: Also why has this not borked the tests :-S

ashb: Oh.

ashb: Probably cos this was only 'tested' from the process function which wasn't tested well enough. I'll add more

ashb:
@mattmb Fixed the app/add typo and added tests. We've just released 0.5.1 to pypi so might be worth merging this today/tomorrow and dealing with any fallout then calling this 0.6.0

PR #105, opened 2015-05-26
Support multiple stacks

mattmb:
This makes it easier to create multiple stacks of the same app and environment. It adds a random string after the stack name so that it does not clash, this is associated with a tag, such as inactive or active.

This doesn't solve the whole problem because some resources still need renaming (i.e. dns-names, S3 buckets). However you no longer need to rename all references to an environment e.g. renaming prod to prod1.

A change is also needed to bootstrap-salt (PR to follow).

I _think_ this is backwards compatible. But it is more code than I wanted to write...

Needs some docs.

### DIFF
--- a/bootstrap_cfn/fab_tasks.py
+++ b/bootstrap_cfn/fab_tasks.py
@@ -169,43 +169,6 @@ def user(username):
     env.user = username
 
 
-def generate_stack_name():
-    """
-    Used to generate new stack name.
-
-    Format <application>-<environment>-<tag>
-
-    tag is optional and defaults to 'active'
-    """
-    if hasattr(env, 'tag'):
-        tag = env.tag
-    else:
-        tag = 'active'
-        env.tag = tag
-    legacy_name = "{0}-{1}".format(env.application, env.environment)
-    cfn_config = get_config()
-    try:
-        r53_conn = get_connection(R53)
-        zone_name = cfn_config.data['master_zone']
-        zone_id = r53_conn.get_hosted_zone_id(zone_name)
-        record_name = "stack.{0}.{1}".format(tag, legacy_name)
-        stack_suffix = uuid.uuid4().__str__()[-8:]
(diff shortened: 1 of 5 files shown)


bootstrap_cfn/fab_tasks.py
```diff
@@ -148,16 +169,111 @@ def user(username):
...
+    if hasattr(env, 'tag'):
+        tag = env.tag
+    else:
+        tag = 'active'
+        env.tag = tag
+    legacy_name = "{0}-{1}".format(env.application, env.environment)
+    cfn_config = get_config()
+    try:
+        r53_conn = get_connection(R53)
+        zone_name = cfn_config.data['master_zone']
```
koikonom: What if I don't use a master zone? Isn't elb:hosted_zone the same ?

mattmb: Looks like I've covered that by catching the KeyError and returning the legacy_name?

bootstrap_cfn/fab_tasks.py
```diff
@@ -148,16 +169,111 @@ def user(username):
...
+    stack_suffix1 = r53_conn.get_record(zone_name, zone_id, record1, 'TXT')
+    stack_suffix2 = r53_conn.get_record(zone_name, zone_id, record2, 'TXT')
+    fqdn1 = "{0}.{1}".format(record1, zone_name)
+    fqdn2 = "{0}.{1}".format(record2, zone_name)
+    r53_conn.update_dns_record(zone_id, fqdn1, 'TXT', '"{0}"'.format(stack_suffix2))
+    r53_conn.update_dns_record(zone_id, fqdn2, 'TXT', '"{0}"'.format(stack_suffix1))
+
+
 def get_stack_name():
     """
```
koikonom, line 196: As discussed, this and generate_stack_name should be one method.

PR #106, opened 2015-06-01
Fix Remote Deletion of SSL Certificates

sevenmachines:
get_remote_certificate doesnt need ssl data passed anymore,
this makes sure that no more calls pass this argument

Closes (but only once we release a version) ministryofjustice/template-deploy#62


ashb:
Changelog? 

Tests so it doesn't break without us noticing again?

sevenmachines:
Added changelog and unit test tweaks

PR #109, opened 2015-06-12
Return the same ELB name as the one that's created

ciprianc:
Right now when we name an elb like: "something-something", it ends up in cloudformation as "somethingsomething", but when we transform that into the pillar data it's still kept using the version with a dash.
That makes deployments not work, as they call the API using the name which has a dash, instead of the renamed version that ended up being created.


ciprianc:
Closing this as we agreed to do it in container.sls rather than here.

PR #111, opened 2015-06-14
Print additional values from the exception object when cfn_create fails.

### LINKED ISSUE #110: When the s3 policy json file is missing, cfn_create dies without message
Example 1 using current bootstrap_cfn:

```
21:20 $ fab application:crimebillingonline aws:cbo environment:opstest1 config:./cloudformation/crimebillingonline.yaml passwords:./cloudformation/crimebillingonline-secrets.yaml cfn_create

Fatal error: Failed to create:

Aborting.
Failed to create:
```

Example 2: using a local copy of bootstrap_cfn modified to print the message:

```
21:28 $ fab application:crimebillingonline aws:cbo environment:opstest1 config:./cloudformation/crimebillingonline.yaml passwords:./cloudformation/crimebillingonline-secrets.yaml cfn_create

Fatal error: Failed to create: [Errno 2] No such file or directory: './cloudformation/s3-bucket-policy-opstest1.json'

Aborting.
Failed to create: [Errno 2] No such file or directory: './cloudformation/s3-bucket-policy-opstest1.json'
```

fab_tasks.py
Line: 345

```
    try:
        stack = cfn.create(stack_name, cfn_config.process())
    except Exception as e:
        abort(red("Failed to create: {error}".format(error=e.message)))
```

to make it print the error message:

```
    try:
        stack = cfn.create(stack_name, cfn_config.process())
    except Exception as e:
        abort(red("Failed to create: errno: {0} message: {1} strerror: {2}".format(e.errno, e.message, e.strerror)))
```

The problem is that in this case, message is empty, while errno and strerror contain values.
The above seems a reasonable workaround that covers all cases as it

bennythejudge:
If cfn_create fails, right now the code only prints the message value, which is sometimes empty,
  while strerror, errno and filename contain useful information.
  The empty message can be confusing and require time to troubleshoot.
  This PR to fix ministryofjustice/bootstrap-cfn#110

### DIFF
--- a/bootstrap_cfn/fab_tasks.py
+++ b/bootstrap_cfn/fab_tasks.py
@@ -341,8 +341,9 @@ def cfn_create():
     # Inject security groups in stack template and create stacks.
     try:
         stack = cfn.create(stack_name, cfn_config.process())
-    except Exception as e:
-        abort(red("Failed to create: {error}".format(error=e.message)))
+    except:
+        import traceback
+        abort(red("Failed to create: {error}".format(error=traceback.format_exc())))
 
     print green("\nSTACK {0} CREATING...\n").format(stack_name)
 


ashb:
Can you provide an example of the before and after?

bennythejudge:
There is an example of before and after within the issue:
https://github.com/ministryofjustice/bootstrap-cfn/issues/110

ashb:
Can you see what this produces?

```
import traceback
abort(red("Failed to create: {error}".format(error=traceback.format_exc())))
```

PR #117, opened 2015-06-22
Automatic tagging of cloudformation and resources  with environment name

sevenmachines:
By passing tags while creating our cloudformation stack we can
propagate these tags to our created resources, enabling easy
separation of environments for billing purposes

### DIFF
--- a/bootstrap_cfn/cloudformation.py
+++ b/bootstrap_cfn/cloudformation.py
@@ -17,10 +17,11 @@ class Cloudformation:
         self.aws_region_name = aws_region_name
         self.conn_cfn = utils.connect_to_aws(boto.cloudformation, self)
 
-    def create(self, stack_name, template_body):
+    def create(self, stack_name, template_body, tags):
         stack = self.conn_cfn.create_stack(stack_name=stack_name,
                                            template_body=template_body,
-                                           capabilities=['CAPABILITY_IAM'])
+                                           capabilities=['CAPABILITY_IAM'],
+                                           tags=tags)
         return stack
 
     def delete(self, stack_name):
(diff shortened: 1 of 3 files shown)


sevenmachines:
TAGGED RESOURCES LIST

RDS Subnet Group
VPC Internet Gateway
VPC Route Table
VPC VPC
EC2 Load Balancer
RDS DB Instance
VPC Subnet
EC2 Instance
EC2 Security Group
S3 Bucket

setup.py
```diff
@@ -25,6 +25,8 @@
     setup_requires=[
         'mock>=1.0.1',
         'testfixtures>=4.1.2',
+        'troposphere>=1.0.0',
```
ashb: Given troposphere is already in the install_requires list we shouldn't need to add it here. What does this fix or what breaks without it?

sevenmachines: Its inclusion was a mistake, I'll sort and re-PR tomorrow

On 22 June 2015 17:30:31 BST, Ash Berlin notifications@github.com wrote:

> > @@ -25,6 +25,8 @@
> >      setup_requires=[
> >          'mock>=1.0.1',
> >          'testfixtures>=4.1.2',
> > -        'troposphere>=1.0.0',
> 
> Given troposphere is already in the install_requires list we shouldn't
> need to add it here. What does this fix or what breaks without it?
> 
> ---
> 
> Reply to this email directly or view it on GitHub:
> https://github.com/ministryofjustice/bootstrap-cfn/pull/117/files#r32953939

## 

Sent from my Android device with K-9 Mail. Please excuse my brevity.

sevenmachines:
Updated

PR #118, opened 2015-06-22
Deprecate ec2.tags.Env and warn about its use

sevenmachines:
The Env tag is now deprecated for specifying the environment on the
command line. This change will ensure we ignore it and warn about
its use. ~~(closes ministryofjustice/template-deploy # 51)~~


ashb:
#51 is about the container versions and which containers to deploy to hosts so I think this is a different issue.

ashb:
Not to be merged before #117

Issue #120, opened 2015-06-23
Pin down the Cipher policy for ELB listener on port 443


bennythejudge:
This refers to an [existing ticket](https://github.com/ministryofjustice/platforms-team/issues/8) on Waffle.
After investigations, the current Amazon default policy ("PolicyName": "ELBSecurityPolicy-2015-05") is acceptable and works with IE7.
Pin down this policy in cloudformation via troposphere in config.py.
More information can be found [here](http://docs.aws.amazon.com/ElasticLoadBalancing/latest/DeveloperGuide/elb-security-policy-table.html)

bennythejudge:
1) Amazon security policy for ELB Listener can be selected via YAML:

```
  elb:
    - name: opstest1
      hosted_zone: advocatedefencepayments.dsd.io.
      scheme: internet-facing
      certificate_name: opstest1-cert
      listeners:
        - LoadBalancerPort: 80
          InstancePort: 80
          Protocol: HTTP
        - LoadBalancerPort: 443
          InstancePort: 80
          Protocol: HTTPS
          PolicyNames:
            - ELBSecurityPolicy-2015-05
```

I have tested against current version of bootstrap-cfn (0.5.0) and it works. 

2) the above works only with current policy (2015-05), while it doesn't work with previous (older) versions.
Trying to use 2015-03 results in cfn_create error with the following message:

```
There is no policy with name ELBSecurityPolicy-2015-03 for load balancer ELB-opstest1
```

3) testing if [this](https://forums.aws.amazon.com/thread.jspa?messageID=595522#595522) is a solution.

PR #124, opened 2015-06-27
Pindown the SSL cipher policy in CFN

bennythejudge:
Add the current SSL cipher policy (Amazon default) via troposphere
  Pindown a policy either in the CFN YAML file or as an implicit default via troposphere
  Value of property PolicyNames must be of type List of String
  Updated tests
  Updated CHANGELOG and README

Fixes ministryofjustice/bootstrap-cfn#120


README.rst
```diff
@@ -147,3 +147,52 @@ You can enable encryption for your DB by adding the following::
...
+		    listeners:
+		      - LoadBalancerPort: 80
+		        InstancePort: 80
+		        Protocol: HTTP
+		      - LoadBalancerPort: 443
+		        InstancePort: 80
+		        Protocol: HTTPS
+		        # following can be commented
+		        PolicyNames:
+		         - PinDownSSLNegotiationPolicy201505
```
ashb: It's not clear that this example is showing, or what purpose it serves. Can you explain your thinking here?

README.rst
```diff
@@ -147,3 +147,52 @@ You can enable encryption for your DB by adding the following::
...
+
+To change the policy edit file::
+
+		bootstrap-cfn/bootstrap_cfn/config.py
```
ashb: This needs to be re-worded slightly - make it clear that if you need a different SSL polcy you will have to open a PR against this repo for instance - i.e. projects can't just 'edit' this file.

bootstrap_cfn/config.py
```diff
@@ -25,7 +25,11 @@ class ProjectConfig:
     config = None
 
     def __init__(self, config, environment, passwords=None):
-        self.config = self.load_yaml(config)[environment]
+        try:
+            self.config = self.load_yaml(config)[environment]
+        except:
```
ashb: This `except` is too generic - we should only catch `KeyError` here

bootstrap_cfn/config.py
```diff
@@ -426,6 +436,11 @@ def elb(self):
                         ":server-certificate/",
                         "{0}-{1}".format(cert_name, self.stack_name)]
                     )
+                    # if not present, add the default cipher policy
+                    if 'PolicyNames' not in listener:
+                        print "\n\n[INFO] ELB Listener for port 443 has no SSL Policy. " + \
```
ashb: `logging.debug` instead of `print` please

PR #126, opened 2015-07-03
Add function to delete DNS records

mattmb:
Adds a function to delete DNS records. To be consumed by other packages that use the create functions here.


mattmb:
Don't strictly need it. I remember why I decided not to tidy up dns records, it's because you'd have to do something like fab salt_delete cfn_delete, instead of just cfn_delete.

ashb:
Going to close without merge then if it's alright? bootstrap-cfn isn't meant to be a general AWS toolkit and it's in danger of becoming that if we add things like this.

PR #132, opened 2015-07-20
Use automatic resource naming for S3 buckets

### LINKED ISSUE #123: Let Cloudformation generate names for S3 resources
We currently hard code the name of the ELB, RDS and S3 buckets that bootstrap-cfn creates.

This means that creating duplicate environments of the same flavour (i.e. another 'dev' env) means we have to duplicate the block in the YAML and change some identifiers.

Instead we can do this:

``` json
{
    "Outputs": {
      "staticbucket": {
        "Description": "x",
        "Value": { "Ref": "StaticBucket" }
      }
    },
    "Resources": {
        "StaticBucket": {
            "Properties": {
                "AccessControl": "BucketOwnerFullControl"
            },
            "Type": "AWS::S3::Bucket"
        }
    }
}
```

Note how there's no `BucketName` specified.

I went and created this template and cloudformation with a stack name of "Ash-s3-test" and the bucket that was created was `ash-s3-test-staticbucket-1hgn7cc7bhnak` – that's `<stackname>-<logical-resource-name>-<random>`. Given that we can make it an output this means we can find the name given in the same way we do with RDS.

We should do similar things for the ELB (where the name of ELB doesn't matter, just the DNS name) and RDS (which we don't need the name at all anyway - just the endpoint which is already an output)

sevenmachines:
Currently we use hard coded names for resources like S3. This change
will allow us to use dynamic names auto-generated by aws and exposed
as an output. (Closes ministryofjustice/bootstrap-cfn#123)
- S3 bucket names are optional in config, default to dynamic ones
- Add a method and fab_file function to allow getting a filtered
  list of resources
- Update tests to look for new BucketName output and policy changes
- PEP8 fixes

### DIFF
--- a/bootstrap_cfn/cloudformation.py
+++ b/bootstrap_cfn/cloudformation.py
@@ -58,11 +58,28 @@ class Cloudformation:
             load_balancers: Set of stack resources containing only
                 load balancers for this stack
         """
+        resource_type = 'AWS::ElasticLoadBalancing::LoadBalancer'
+        return self.get_resource_type(stack_name_or_id, resource_type)
+
+    def get_resource_type(self, stack_name_or_id, resource_type=None):
+        """
+        Collect up a set of specific stack resources
+
+        Args:
+            stack_name_or_id (string): Name or id used to identify the stack
+            resource_type(string): The resource type identifier
+
+        Returns:
+            resources: Set of stack resources containing only
+                the resource type for this stack
+        """
         # get the stack
-        load_balancers = []
+        resources = []
         stack = self.conn_cfn.describe_stacks(stack_name_or_id)
         if stack:
-            fn = lambda x: x.resource_type == 'AWS::ElasticLoadBalancing::LoadBalancer'
-            # get the load balancers group
-            load_balancers = filter(fn, stack[0].list_resources())
-        return load_balancers
+            resources = stack[0].list_resources()
+            if resource_type:
+                # get the resources
+                resources = filter(lambda x: x.resource_type == resource_type,
+                                   resources)
+        return resources
--- a/bootstrap_cfn/config.py
+++ b/bootstrap_cfn/config.py
@@ -88,13 +88,7 @@ class ConfigParser:
             ))
 
         if 's3' in self.data:
-            s3 = self.s3()
-            map(template.add_resource, s3)
-            template.add_output(Output(
-                "StaticBucketName",
-                Description="S3 bucket name",
-                Value=Ref("StaticBucket")
-            ))
+            self.s3(template)
 
         template = json.loads(template.to_json())
         if 'includes' in self.data:
@@ -266,7 +260,17 @@ class ConfigParser:
         # return json.loads(json.dumps(dict((r.title, r) for r in resources), cls=awsencode))
         return resources
 
-    def s3(self):
+    def s3(self, template):
+        """
+        Create an s3 resource configuration from the config file data.
+        This will produce Bucket and BucketPolicy resources along with
+        the bucket name as output, these are all added to the troposphere
+        template.
+
+        Args:
+            template:
+                The cloudformation template file
+        """
         # As there are no required fields, although we may not have any
         # subkeys we still need to be able to have a parent key 's3:' to
         # signify that we want to create an s3 bucket. In this case we
@@ -293,7 +297,7 @@ class ConfigParser:
             arn = Join("", ["arn:aws:s3:::", Ref(bucket), "/*"])
             policy = {
                 'Action': ['s3:GetObject'],
-                "Resource": [arn],
+                "Resource": arn,
                 'Effect': 'Allow',
                 'Principal': '*'}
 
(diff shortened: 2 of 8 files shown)


bootstrap_cfn/cloudformation.py
```diff
@@ -58,11 +58,30 @@ def get_stack_load_balancers(self, stack_name_or_id):
...
-            fn = lambda x: x.resource_type == 'AWS::ElasticLoadBalancing::LoadBalancer'
-            # get the load balancers group
-            load_balancers = filter(fn, stack[0].list_resources())
-        return load_balancers
+            if resource_type:
+                fn = lambda x: x.resource_type == resource_type
+                # get the resources
+                resources = filter(fn, stack[0].list_resources())
+            else:
+                resources = stack[0].list_resources()
```
ashb: This could be written a bit clearer. In pseudo code:

```
resources = stack.list_resources()
if resource_type:
  resources = apply filter(resources)
```

The rule (which is never a hard and fast rule, cos they never are) to apply is that if you can easily limit if/else then it's easier to follow the code path through.

bootstrap_cfn/config.py
```diff
@@ -264,29 +269,46 @@ def iam(self):
...
+        # subkeys we still need to be able to have a parent key 's3:' to
+        # signify that we want to create an s3 bucket. In this case we
+        # set up an empty (no options set) dictionary
+        if isinstance(self.data['s3'], dict):
+            present_keys = self.data['s3'].keys()
+            # TEST FOR REQUIRED FIELDS AND EXIT IF MISSING ANY
+            for i in required_fields.keys():
+                if i not in present_keys:
+                    print "\n\n[ERROR] Missing S3 fields [%s]" % i
+                    sys.exit(1)
```
ashb: Since we no longer have any require keys we should remove the loop/check too.

bootstrap_cfn/config.py
```diff
@@ -264,29 +269,46 @@ def iam(self):
...
+                bucket = Bucket(
+                    "StaticBucket",
+                    AccessControl="BucketOwnerFullControl",
+                    BucketName=self.data['s3']['static-bucket-name'],
+                )
+        else:
+            bucket = Bucket(
+                "StaticBucket",
+                AccessControl="BucketOwnerFullControl",
+            )
```
ashb: You can simplify this too - you can set properties on the troposphere instances:

``` python
bucket = Bucket(
    "StaticBucket",
    AccessControl="BucketOwnerFullControl",
)

if 'static-bucket-name' in present_keys:
    bucket.BucketName = self.data['s3']['static-bucket-name']
```

ashb:
Apart from the minor fixes mentioned in comments the change looks good, but this is only part of the changes needed for #123 - we also need to adjust ELBs and RDS instances similarly.

I'm happy to merge this change before then but we need to change the PR and commit message to not close the issue as it's not complete yet.

sevenmachines:
Fixup commit for review.
Note, I've seperated out the ELB and RDS equivalents into their own issues so closing this one is ok

sevenmachines:
Have squashed and rebased

bootstrap_cfn/config.py
```diff
@@ -90,6 +90,11 @@ def process(self):
         if 's3' in self.data:
             s3 = self.s3()
             map(template.add_resource, s3)
+            template.add_output(Output(
+                "StaticBucketName",
+                Description="S3 bucket name",
+                Value=Ref("StaticBucket")
+            ))
```
ashb: Ok time to refactor here - we shouldn't be creating things outside of the `s3` function that relate to things inside it.

Options I can think of:
- lists-of-lists from `s3()`
- we could pass the Troposphere template object in to the class and let the s3 call `add_resource` or `add_output` as it needs.
- We can return a list of objects and detect where to add them. Something like:
  
  ``` python
      objs = self.s3()
      self.add_to_template(template, objs)
  
  #  …
  
  def add_to_template(self, template, objects)
      for object in objs:
          if isinstance(obj, troposphere.Output):
              template.add_output(obj)
            elif isinstance(obj, troposphere.Parameter):
              template.add_parameter(obj)
  ```

bootstrap_cfn/config.py
```diff
@@ -262,31 +267,33 @@ def iam(self):
...
+        # a reasonable default of public 'Get' access
         if 'policy' in present_keys:
             policy = json.loads(open(self.data['s3']['policy']).read())
         else:
-            arn = 'arn:aws:s3:::%s/*' % self.data['s3']['static-bucket-name']
+            arn = Join("", ["arn:aws:s3:::", Ref(bucket), "/*"])
             policy = {
                 'Action': ['s3:GetObject'],
-                'Resource': arn,
+                "Resource": [arn],
```
ashb: Does this really need to be an array now?

tests/tests.py
```diff
@@ -148,16 +183,16 @@ def test_custom_s3_policy(self):
...
+                'Effect': 'Allow',
+                'Principal': {'AWS': '*'}
             }
         ]
 
         project_config = ProjectConfig('tests/sample-project.yaml', 'dev')
 
         project_config.config['s3'] = {
-            'static-bucket-name': 'moj-test-dev-static',
+            'StaticBucketName': 'moj-test-dev-static',
```
ashb: I don't think this should have changed - this is the input yaml format which should remain as it was.

bootstrap_cfn/config.py
sevenmachines: I've used the 'returning the output list' option with ELB but I'm moving more towards the 'passing the template' option

sevenmachines:
Fixups added,
- pass template to s3
- [arn] array change
- Update to tests to handle template being passed to s3
- static-bucket-name test change

sevenmachines:
Updated testing to cover s3 outputs values as well

PR #133, opened 2015-07-20
autoscale class that can find the scaling group based on a name and s…

stuartornum:
…et tags against it

### DIFF
--- a/tests/test_autoscale.py
+++ b/tests/test_autoscale.py
@@ -1,0 +1,74 @@
+import os
+import tempfile
+import unittest
+
+import boto.ec2.autoscale
+from boto.ec2.autoscale import AutoScaleConnection
+from boto.ec2.autoscale.group import AutoScalingGroup
+
+import mock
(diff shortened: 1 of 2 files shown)


stuartornum:
This passes if I run it locally...

tests/test_autoscale.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def test_loaded(self):
+        pass
+
+    def test_set_autoscaling_group(self):
+        with mock.patch(
+                'boto.ec2.autoscale.AutoScaleConnection.get_all_groups',
+                get_all_groups):
+            with mock.patch(
+                    'boto.ec2.autoscale.connect_to_region', connect_to_region):
```
andyhd: Change this to:
```python
with mock.patch('boto.ec2.autoscale', mock_autoscale):
    mock_autoscale.connect_to_region.side_effect = connect_to_region
```
That might work

PR #134, opened 2015-07-20
Add Boto3 in addition to Boto

pwyborn:
@mattmb Add Boto3 Can you please check - I am assuming that for the time being the boto config stays the same? 

(Boto3, the next version of Boto, is now stable and recommended for general use. It can be used side-by-side with Boto in the same project....) see https://waffle.io/ministryofjustice/platforms-team/cards/55a587113045614d0001e750.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,22 @@
-## Version 0.5.x
+## Version 0.5.6
+* Automaticly generate the RDS identifier
+
+## Version 0.5.5
+* Make it possible to override the ConfigParser so that sub-modules can update the CloudFormation config.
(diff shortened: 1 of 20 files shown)


ashb:
@pwyborn We need to do more than include the boto3 module for this card - if only it were that easy. We need to start switching the API calls we use from boto2 to boto3. Sorry I didn't quite make that clear in the original ticket.

sevenmachines:
placeholder link for the boto3 migration guide...
http://boto3.readthedocs.org/en/latest/guide/index.html#migration-guides

bennythejudge:
closing as this is a duplicate

PR #138, opened 2015-07-22
(INVALID) ELB Automatic resource naming

sevenmachines:
This change removes settings LoadBalancerName manually and
uses the AWS auto naming instead. ELB names are now not
mandatory. (Closes ministryofjustice/bootstrap-cfn#136)
- Add elb dns names as outputs
- Arbitrary elb logical ids
- Remove default security group for ELBs
- Pass troposphere template to elb function so that resources
  and outputs can be set up internal to the function
- Check we have AutoScalingGroups before attaching to them
- Update tests for elb method requiring template
- Merge test:test_elb security group checks into overall cfg check


sevenmachines:
Needs testing

bootstrap_cfn/config.py
```diff
@@ -375,32 +372,52 @@ def rds(self):
...
             'scheme': 'Scheme',
-            'name': 'LoadBalancerName',
             'hosted_zone': 'HostedZoneName'
         }
 
         elb_list = []
         elb_sgs = []
         # COULD HAVE MULTIPLE ELB'S (PUBLIC / PRIVATE etc)
+        elb_count = 0
         for elb in self.data['elb']:
```
ashb, line 396: We can do this in a more pythonic manner:

```
In [5]: for a,b in enumerate(x):
    print repr(a), repr(b)
   ...:
0 'a'
1 'b'
2 'c'
```

ashb:
I think we're doing this wrong.

We can still specify a logical/resource name and thus make name a required field. We just shouldn't feed that in to the LoadBalancerName field and let AWS generate _that_. This means we don't have the problems with ministryofjustice/template-deploy#120

sevenmachines:
Dropping this for https://github.com/ministryofjustice/bootstrap-cfn/pull/143

PR #139, opened 2015-07-23
Added display_elb_dns_entries task

bennythejudge:
- for a given environment, for each ELB display the corresponding DNS entry
- updated CHANGELOG
- added tests

### DIFF
--- a/bootstrap_cfn/elb.py
+++ b/bootstrap_cfn/elb.py
@@ -3,7 +3,7 @@ import logging
 import boto.ec2.elb
 
 from bootstrap_cfn import cloudformation, iam, utils
-from bootstrap_cfn.errors import CloudResourceNotFoundError
+from bootstrap_cfn.errors import BootstrapCfnError, CloudResourceNotFoundError
 
 
 class ELB:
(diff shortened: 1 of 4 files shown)


bootstrap_cfn/elb.py
```diff
@@ -84,3 +84,31 @@ def set_ssl_certificates(self, ssl_config, stack_name):
...
+        Returns:
+            list of dict: [{'elb_name': string, 'dns_name': string}]
+
+        Raises:
+            The boto call raises self.ResponseError(response.status, response.reason, body)
+            if the AWS call returns an error, for example if the ELB name does not exist
+        """
+        lb_name_dns = []
+        load_balancer_resources = self.cfn.get_stack_load_balancers(stack_name)
+        for l in load_balancer_resources:
```
ashb: Can this for loop be simplified to something like:

``` python
lb_ids = [l.physical_resource_id for l in load_balancer_resources]
lb_details = self.conn_elb.get_all_load_balancers(load_balancer_names=lb_ids)
```

We can then check that the number of results we get back is the same as the number of ids we passed.

bootstrap_cfn/fab_tasks.py
```diff
@@ -420,3 +420,25 @@ def get_cloudformation_tags():
...
+    """
+    Prints out the ELB name(s) and the corresponding DNS name(s) for every ELB
+    in the environment provided.
+    """
+    stack_name = get_stack_name()
+    if not stack_name:
+        abort("Could not retrieve stack_name")
+    elb = get_connection(ELB)
+    if not elb:
+        abort("Could not retrieve elb object")
```
ashb: This check isn't needed - get_connection will thrown an exception if something goes wrong.

bootstrap_cfn/fab_tasks.py
```diff
@@ -420,3 +420,25 @@ def get_cloudformation_tags():
...
+    elb_dns_list = elb.list_domain_names(stack_name)
+    import json
```
ashb: Is there a reason we are displaying this as json? Is it intended for consumption by a program - if not maybe it should be in plain text/some kind of table.

bootstrap_cfn/fab_tasks.py
```diff
@@ -420,3 +420,25 @@ def get_cloudformation_tags():
...
+    for elb_dns in elb_dns_list:
+        print json.dumps({
+            'elb_name': elb_dns['elb_name'],
+            'dns_name': elb_dns['dns_name']},
+            sort_keys=True, indent=4, separators=(',', ': '))
```
ashb: This could be simplifeid to 

``` python
print json.dumps(elb_dns, sort_keys=True, indent=4, separators=(',', ': '))
```

bootstrap_cfn/fab_tasks.py
```diff
@@ -420,3 +420,25 @@ def get_cloudformation_tags():
...
+    return
```
ashb: This return is un-needed and isn't very pythonic

bootstrap_cfn/fab_tasks.py
bennythejudge: I thought formatted json would be humanly readable and it could be used down the chain

PR #140, opened 2015-07-23
Make config parser extensible

mattmb:
This makes it possible to subclass the config parser, override it's methods and then use this subclass when generating the CloudFormation json.

We need this so that bootstrap-salt can update the user data section of the launch configuration.


sevenmachines:
Needs some conflicts sorted out :(

mattmb:
Good news, just the changelog :smile:

sevenmachines:
always nice when that happens :)

PR #141, opened 2015-07-23
Use automatic naming of the RDS

sevenmachines:
This removes the use of the RDS identifier setting.
- Pass template to config:rds method
- Randomise a DBInstanceIdentifierName
- Update tests to use template passed to rds()
- Changed 2 test comparisons to standard expected_data, actual_data order
- Drop mention of identifier from sample project

(Closes ministryofjustice/bootstrap-cfn#137)

### DIFF
--- a/bootstrap_cfn/config.py
+++ b/bootstrap_cfn/config.py
@@ -3,7 +3,6 @@ import logging
 import os
 import sys
 import textwrap
-import uuid
 
 from troposphere import Base64, FindInMap, GetAZs, GetAtt, Join, Output, Ref, Tags, Template
 from troposphere.autoscaling import AutoScalingGroup, BlockDeviceMapping, \
@@ -335,6 +334,7 @@ class ConfigParser(object):
 
         optional_fields = {
             'storage-encrypted': 'StorageEncrypted',
+            'identifier': 'DBInstanceIdentifier'
         }
 
         # LOAD STACK TEMPLATE
@@ -363,10 +363,6 @@ class ConfigParser(object):
         )
         resources.append(database_sg)
 
-        identifier = ("RDS%s%s"
-                      % ((self.stack_name).replace('-', '').replace('.', '').replace('_', ''),
-                         uuid.uuid4().__str__()[-8:]))
-
         rds_instance = DBInstance(
             "RDSInstance",
             PubliclyAccessible=False,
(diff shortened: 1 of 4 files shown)


mattmb:
Hmmm, I think it's the identifier that's most problematic?

Maybe we should have random DBName too. But it's the identifier that has to be unique.

bootstrap_cfn/config.py
```diff
@@ -356,6 +357,7 @@ def rds(self):
             AutoMinorVersionUpgrade=False,
             VPCSecurityGroups=[GetAtt(database_sg, "GroupId")],
             DBSubnetGroupName=Ref(rds_subnet_group),
+            DBInstanceIdentifier=identifier,
```
ashb: I don't think we should need to specify this. From http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-rds-database-instance.html#cfn-rds-dbinstance-dbinstanceidentifier

> **DBInstanceIdentifier**
> 
>    A name for the DB instance. If you don't specify a name, AWS CloudFormation generates a unique physical ID and uses that ID for the DB instance. For more information, see Name Type.
> 
>    _Required: No_

tests/tests.py
```diff
@@ -564,7 +592,7 @@ def test_process_with_vpc_config(self):
         wanted = ["dbhost", "dbport"]
         output_names = cfn_template['Outputs'].keys()
         output_names.sort()
-        compare(output_names, wanted)
+        compare(wanted, output_names)
```
ashb, line 681: I'm not sure if there's a proper order for wanted vs expected (the docs just say one and two I think) but why did we swap these parameters around?

bootstrap_cfn/config.py
```diff
@@ -312,7 +310,6 @@ def rds(self):
             'backup-retention-period': 'BackupRetentionPeriod',
             'db-master-username': 'MasterUsername',
             'db-master-password': 'MasterUserPassword',
-            'identifier': 'DBInstanceIdentifier',
```
ashb, line 332: To make back-compat/upgrade easy could we make this an optional field?

sevenmachines, line 332: will do

bootstrap_cfn/config.py
sevenmachines: Can do this as part of the above change too

tests/tests.py
sevenmachines, line 681: Not proper so much as consistent, mosts of the tests run an expected/actual comparison order, best to keep them all the same if possible

PR #149, opened 2015-08-20
Check for AWS_ROLE_ARN_ID in utils:connect_to_aws before assuming role

sevenmachines:
connect_to_aws checks for the cross-account profile and then tries to
assume a role based on that. This means that if are not using the
cross-account as a profile name we will skip assuming roles. This change
will also attempt to assume a role if we have a AWS_ROLE_ARN_ID variable
set.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## UNRELEASED
+
+* Change AWS connection to use STS AssumeRole when AWS_ROLE_ARN_ID
+environment variable is specified in addition to just the profile being
+called "cross-account".
+
 ## Version 0.5.9
 
 * Bump version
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## UNRELEASED
+
+* Check for AWS_ROLE_ARN_ID in utils:connect_to_aws before attempting to
+assume a role using the STS service
```
ashb: Could you rephrase this to say something like "Change AWS connection to use STS AssumeRole when AWS_ROLE_ARN_ID environment variable is specified in addition to just the profile being called "cross-account".

sevenmachines: Fixup added

ashb:
Thanks. Rebase and merge at your leisure. Do we know if this actually fixed our problem on LAALAA?

sevenmachines:
@ashb I tested it out on laalaa-DEPLOY first, it works now.

PR #152, opened 2015-09-09
Added elasticache with engine redis

sevenmachines:
Can now add an elasticache section to the template to get a redis
based elasticache. The number of clusters in the replication group,
node_type and port can be specified. A list of redis backup files
can be provided also, to seed the new elasticache resource.
The replication groups name and engine is provided as an output.

``` yaml
elasticache:
  clusters: 3
  node_type: cache.m1.small
  port: 10024
  seeds:
    s3:
    - test-redis-bucket-947923urhiuy8923d/redis.rdb
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -54,7 +54,7 @@ If your ``$CWD`` is anywhere else, you need to pass in a path to particular fabr
 - **config:/path/to/file.yaml** - The location to the project YAML file
 
 Multiple Stacks
-=================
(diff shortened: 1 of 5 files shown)


niallcreech:
Still need to...
* rename  rds_prop in config check, maybe move into a helper function. 
* Test that our security groups are actually right and instances can get to endpoints

PR #153, opened 2015-09-09
Permit EC2 machines to tag EBS volumes

mattpep:
The EC2 instance will be tagged automatically by virtue of being part of
the Auto Scaling Group.  But due to a (possibly intentional) quirk of
AWS, EBS volumes created automatically for an EC2 instance do not get
tagged (even though EC2 instances do).

Instead, we rely on salt to tag the volume since salt will run when the
ec2 instance first comes up.

Note that EBS volumes are considered EC2 entities for the sake of IAM
role/permission syntax.


niallcreech:
You'll need to look over the failing travis checks, maybe run the nosetests...

mattpep:
tests now written - if good I'll squish prior to a merge.

PR #159, opened 2015-10-29
Feature maintenance mode

filipposc5:
- Adds enter_maintenance and exit_maintenance fab tasks
- Adds support for updating route53 dns aliases

### DIFF
--- a/bootstrap_cfn/fab_tasks.py
+++ b/bootstrap_cfn/fab_tasks.py
@@ -15,6 +15,8 @@ from fabric.utils import abort
 from bootstrap_cfn.cloudformation import Cloudformation
 from bootstrap_cfn.config import ConfigParser, ProjectConfig
 from bootstrap_cfn.elb import ELB
+from bootstrap_cfn.errors import BootstrapCfnError
+from bootstrap_cfn.errors import CloudResourceNotFoundError
 from bootstrap_cfn.iam import IAM
 from bootstrap_cfn.r53 import R53
 from bootstrap_cfn.utils import tail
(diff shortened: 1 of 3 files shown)


bootstrap_cfn/fab_tasks.py
```diff
@@ -193,6 +193,94 @@ def swap_tags(tag1, tag2):
...
+    """
+    cfn_config = get_config()
+    r53_conn = get_connection(R53)
+
+    cached_zone_ids = {}
+    for elb in cfn_config.data['elb']:
+        if not elb['scheme'] == 'internet-facing':
+            continue
+
+        record = "{0}.{1}".format(elb['name'], elb['hosted_zone'])
```
andyhd: You can write this as:

``` python
record = "{name}.{hosted_zone}".format(**elb)
```

bootstrap_cfn/fab_tasks.py
```diff
@@ -193,6 +193,94 @@ def swap_tags(tag1, tag2):
...
+    elb_conn = get_connection(ELB)
+
+    cfn_config = get_config()
+    stack_name = get_stack_name()
+
+    # get all elbs for this stack from AWS
+    stack_elbs = {}
+    [stack_elbs.update({
+                       getattr(x, 'logical_resource_id'): getattr(x, 'physical_resource_id')
+                       }) for x in elb_conn.cfn.get_stack_load_balancers(stack_name)]
```
andyhd: Rather than creating the dict and then updating it in a list comprehension (which is bad form, btw - this should just be a `for` loop), you can construct the dict with a list comprehension directly. Additionally, I would not use `getattr` here, as it's unnecessarily verbose:

``` python
stack_elbs = dict([
    (x.logical_resource_id, x.physical_resource_id)
    for x in elb_conn.cfn.get_stack_load_balancers(stack_name)])
```

bootstrap_cfn/fab_tasks.py
```diff
@@ -193,6 +193,94 @@ def swap_tags(tag1, tag2):
...
+    Sets internet-facing elbs hostnames
+    back to the ELB DNS alias
+    """
+    r53_conn = get_connection(R53)
+    elb_conn = get_connection(ELB)
+
+    cfn_config = get_config()
+    stack_name = get_stack_name()
+
+    # get all elbs for this stack from AWS
```
andyhd: This comment only explains what the code is doing. It should explain _why_ you're doing it.

bootstrap_cfn/fab_tasks.py
```diff
@@ -193,6 +193,94 @@ def swap_tags(tag1, tag2):
...
+    # filter stack related load balancers (as opposed to all stack elbs in the account)
+    full_load_balancers = elb_conn.conn_elb.get_all_load_balancers(
+        load_balancer_names=stack_elbs.values())
+
+    cached_zone_ids = {}
+    # loop through elb config entries and change internet facing ones
+    for elb in cfn_config.data['elb']:
+        if not elb['scheme'] == 'internet-facing':
+            continue
+        record = "{0}.{1}".format(elb['name'], elb['hosted_zone'])
```
andyhd: Same as before

bootstrap_cfn/fab_tasks.py
```diff
@@ -193,6 +193,94 @@ def swap_tags(tag1, tag2):
...
+        # obtain physical name from dict lookup, by converting elb name into safe name
+        # into logical name
+        phys_name = stack_elbs[get_elb_safe_name(elb['name'])]
+
+        dns_name = [x.dns_name for x in full_load_balancers if x.name == phys_name][0]
```
andyhd: This will fail if the list comprehension is empty. Build the list first, check the length, then access the first element separately.

bootstrap_cfn/fab_tasks.py
```diff
@@ -193,6 +193,94 @@ def swap_tags(tag1, tag2):
...
+    in same zone, so feel free to cache it (and drink a shot because I said 'cache')
+    '''
+    if zone_dict.get(zone_name, False):
+        return zone_dict[zone_name]
+    # not found, look it up, cache it up ..
+    zone_dict[zone_name] = r53_conn.get_hosted_zone_id(zone_name)
+    return zone_dict[zone_name]
+
+
+def get_elb_safe_name(elb_name):
```
andyhd: `get_` should not be used in a function name unless it is an accessor method. This would be better named `escape_elb_name` or `make_safe_elb_name` or similar. Perhaps `slugify` would be better and shorter.

bootstrap_cfn/r53.py
```diff
@@ -5,6 +5,20 @@
 
 class R53:
```
andyhd: This should be `class R53(object):`

bootstrap_cfn/fab_tasks.py
```diff
@@ -193,6 +193,94 @@ def swap_tags(tag1, tag2):
...
+def get_cached_zone_id(r53_conn, zone_dict, zone_name):
+    '''
+    If we are looping through ELBs we may just have different hostnames
+    in same zone, so feel free to cache it (and drink a shot because I said 'cache')
+    '''
+    if zone_dict.get(zone_name, False):
+        return zone_dict[zone_name]
+    # not found, look it up, cache it up ..
+    zone_dict[zone_name] = r53_conn.get_hosted_zone_id(zone_name)
+    return zone_dict[zone_name]
```
andyhd, line 303: Prefer a single exit point. Eg:

``` python
if zone_name not in zone_dict:
    zone_dict[zone_name] = r53_conn.get_hosted_zone_id(zone_name)
return zone_dict[zone_name]
```

andyhd, line 303: Also, this doesn't handle errors from `r53_conn.get_hosted_zone_id()`

bootstrap_cfn/fab_tasks.py
```diff
@@ -193,6 +193,94 @@ def swap_tags(tag1, tag2):
...
+
+
+def get_elb_safe_name(elb_name):
+    '''
+    '''
```
andyhd, line 309: Missing docstring. Does this need to be a top-level function? It's only used once on Line 248, maybe inline it?

filipposc5:
How to test this for your project:  

```
git clone https://github.com/ministryofjustice/bootstrap-cfn
cd bootstrap-cfn 
git checkout feature-maintenance-mode
export P=$PWD
cd <your-project>-deploy
pip uninstall -y bootstrap-cfn
pip install $P
fab environment:.. application:... aws:.. config..etc  enter_maintenance:1.1.1.1
```

Where 1.1.1.1 the IP of your maintenance server.
and to exit 
`fab environment:.. application:... aws:.. config..etc  exit_maintenance`

bootstrap_cfn/fab_tasks.py
filipposc5, line 309: I wish I didn't even have to write it, as it is used in config.py, and in general it is very much tucked away and hidden. I made it top level so that it's more obvious this is happening behind the scenes (as it is happening elsewhere, like in salt content, and I don't know if at some point it breaks backward compatibility with some older stacks). 

I ll call it mold_elb_name .. as it is moldy enough already.

filipposc5:
Still need to add changelog entries and readme but most of @andyhd 's feedback is in (or so I think!)

KrisSaxton:
Confirmed as working on ELK stack in dev environment.  Although curl throws SSL errors unless run in --insecure mode (is this expected)?

filipposc5:
@KrisSaxton I think this is normal because we haven't configured the other endpoint. We should check the whole feature at some point.

bootstrap_cfn/fab_tasks.py
```diff
@@ -193,6 +195,112 @@ def swap_tags(tag1, tag2):
...
+
+    # filter stack related load balancers (as opposed to all stack elbs in the account)
+    full_load_balancers = elb_conn.conn_elb.get_all_load_balancers(
+        load_balancer_names=stack_elbs.values())
+
+    cached_zone_ids = {}
+    # loop through elb config entries and change internet facing ones
+    for elb in cfn_config.data['elb']:
+        if not elb['scheme'] == 'internet-facing':
+            continue
```
andyhd, line 259: Since this loop and filter code is repeated (also on lines 209-211), it should probably be refactored into a function, eg:

``` python
def internet_facing_elbs(cfn_config):
    return filter(lambda elb: elb['scheme'] == 'internet-facing', cfn_config.data['elb'])
```

or even better:

``` python
def is_internet_facing(elb):
    return elb['scheme'] == 'internet-facing'

def internet_facing_elbs(cfn_config):
    return filter(is_internet_facing, cfn_config.data['elb'])

for elb in internet_facing_elbs(cfn_config):
    ...
```

filipposc5:
@andyhd 
Thank you for all the constructive feedback! I 've decided to add and keep just the one function in the spirit of not having too much premature optimisation, as I have a feeling we 'll be revisiting this code in the future.

Feel free to merge when ready.

lukaszraczylo:
Should we provide any 'default failback' ip setting?

PR #163, opened 2015-11-11
Add ability to specify more custom policies for an ELB

sevenmachines:
This change allows us to construct policies to be attached to
an ELB and applied to ports.


lukaszraczylo:
@niallcreech CI has failed & it's still called 'WIP' :)

mattpep:
Looks good, and works too.  (I was in the rare situation of needing the work in this and it was minimal effort to test is for correctness rather than just scanning for legibility). With the 'WIP' disclaimer removed I'd be happy to merge.

PR #165, opened 2015-12-09
Add dynamic VPC address block subnetting

sevenmachines:
This change will mean that there will be an attempt made to
dynamically get unassigned address ranges from within the VPC that
the new VPC will use. The progression is,
- Use VPC subnets from config file if they exist
- If not, try to get an unused address range in the VPC
- If all of the above fail, use the default hardcoded VPC addresses

### DIFF
--- a/bootstrap_cfn/config.py
+++ b/bootstrap_cfn/config.py
@@ -21,7 +21,7 @@ from troposphere.s3 import Bucket, BucketPolicy
 
 import yaml
 
-from bootstrap_cfn import errors, mime_packer, utils
+from bootstrap_cfn import errors, mime_packer, utils, vpc
 
 
 class ProjectConfig:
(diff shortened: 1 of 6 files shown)


sevenmachines:
Seems importing vpc module at the wrong point breaks setuptools creaky testing :( Fixed

mattpep:
This diff looks fine to me, though one thought has come to mind:
This now depends on boto3. I see no removal of boto2; do the two libraries co-exist happily? If not, maybe a line in some release notes about boto3  since it's quite a core library to be silently updating

ltsampros:
@mattpep According to boto3 docs, the two libraries can coexist happily.

sevenmachines:
yes, boto2 and boto3 are happy enough side-by-side. I'm defaulting to boto3 for new stuff as we're migrating to boto3 overall. Its only used in the new code rather than altering the code so I'm not to worry about gradually introducing it

PR #166, opened 2015-12-14
Add vpc peering advanced

sevenmachines:
Add advanced settings and wildcards for peering vpcs

```
* Create new class VPC to handle vpc connections and peering
* Update requirements to include boto3
* Add new fab tasks enable/disable_vpc_peering
* Add boto3 to setup.py requirements
* Add VPC docs page
* Add vpc requester/accepter routes to all route tables in each vpc

* Allow per route table cidr block specification
* Add option to wildcard the entire stack. Making all routes
and cidr ranges peered between the 2 stacks
* Add option to wildcard source or target route tables meaning that
all route tables will have the specified cidr block route applied
* Add option to wildcard a cidr block so that the specified route
table will have all the target stacks cidr block range applied to
it
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -13,6 +13,7 @@ The objective of this repo is to enable MoJ teams to create project infrastructu
 - Elastic Load Balancers (ELB)
 - Relational Database Service (RDS)
 - S3 Storage for web static content
+- `VPC Configuration <docs/vpc-configuration.md>`_
 
 Installation
 ============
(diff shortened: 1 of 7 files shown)


sevenmachines:
Re-testing after a large rebase

sevenmachines:
Rebasing and fixups completed, back in 'Ready'. Note that that means there are fixups to squash before merging

ltsampros:
It looks good. Some minor comments on DRY/style here however I won't block merging for these. Tests seem to run fine. If no one objects, I'll merge this today.

PR #171, opened 2016-01-15
Remove the fallback when route53 stack name is missing

### LINKED ISSUE #170: get_stack_name uses fallback name when route53 lookup fails
get_stack_name uses fallback name of <application>_<environment> when route53 lookup fails. This was done for backwards compatibility reasons. Unfortunately this has the side effect of masking any route53 stack_name lookup error, which all new stacks should have. This seems more damaging and time-consuming than defaulting to failure.

sevenmachines:
get_stack_name uses fallback name of <application>-<environment>
when route53 lookup fails. This was done for backwards compatibility
reasons. Unfortunately, this has the side effect of masking any
route53 stack_name lookup error, which all new stacks should have.

This change will raise errors if the stack name lookup fails
rather than attempt a fallback name, this name is only useful in
the case of very old stacks that created stacks with this style of
name.

(Closes #170)

### DIFF
--- a/bootstrap_cfn/errors.py
+++ b/bootstrap_cfn/errors.py
@@ -34,5 +34,20 @@ class ProfileNotFoundError(BootstrapCfnError):
         )
 
 
+class ZoneIDNotFoundError(BootstrapCfnError):
+    def __init__(self, zone_name):
+        msg = ("Could not find a zone id for zone name '{}'."
+               "Please check that this hosted zone exists "
+               "and is in the account you've specified".format(zone_name))
+        super(ZoneIDNotFoundError, self).__init__(msg)
+
+
+class ZoneRoute53RecordNotFoundError(BootstrapCfnError):
+    def __init__(self, zone_name, zone_id):
+        msg = ("Could not find an AWS Route53 record for zone name '{}' with zone id '{}'. "
+               "Please check that this record exists in the account you've specified".format(zone_name, zone_id))
+        super(ZoneRoute53RecordNotFoundError, self).__init__(msg)
+
+
 class CloudResourceNotFoundError(BootstrapCfnError):
     pass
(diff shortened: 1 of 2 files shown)


sevenmachines:
Success looks something like the below now
    ﻿  
        INFO:bootstrap-cfn:fab_tasks::get_stack_name: Found master zone 'dsd.io' in config...
        INFO:bootstrap-cfn:fab_tasks::get_stack_name: Found zone id 'Z31RX3GZS94JZS' for zone name 'dsd.io'...
        INFO:bootstrap-cfn:fab_tasks::get_stack_name: Found stack suffix 'cc50ef43' for record name 'stack.active.helloworld-dev'...
        INFO:bootstrap-cfn:fab_tasks::get_stack_name: Found stack name 'helloworld-dev-cc50ef43'...

sevenmachines:
bad hosted zone name error

bootstrap_cfn.errors.ZoneIDNotFoundError: Could not find a zone id for zone name 'dssd.io'.Please check that this hosted zone exists and is in the account you've specified

sevenmachines:
Missing master zone in config error

bootstrap_cfn.errors.CfnConfigError: No master_zone in yaml, unable to create/find DNS records for stack name

sevenmachines:
Looking for the stack in the wrong account error

bootstrap_cfn.errors.ZoneIDNotFoundError: Could not find a zone id for zone name 'dsd.io'.Please check that this hosted zone exists and is in the account you've specified

sevenmachines:
Note, I think that breaking back compatibility with stack names is acceptable since anything that has the old stack name system could not possibly work with the current versions of bootstrap cfn anyway. Compatibility was broken many months ago quite early in development. This would still warrant a major version bump though by definition

PR #176, opened 2016-01-22
Separate get and set_stack name

sevenmachines:
The get and set stack name functionality is combined into one
command at the moment. This means that a boto route53 connection
is required, even if we only need to fetch a public address to get
the stack name. This change uses separates the functions so that
dnspythons resolver can be used to fetch the stack name record,
meaning that no AWS credentials are required. set_stack_name uses
the same boto.route53 method as before to create stack name records.

 # Please enter the commit message for your changes. Lines starting


lukaszraczylo:
LGTM.

PR #183, opened 2016-02-15
Add windows support

mattpep:


bootstrap_cfn/config.py
```diff
@@ -91,10 +91,14 @@ def process(self):
     def base_template(self):
         from bootstrap_cfn import vpc
         t = Template()
-
-        t.add_mapping("AWSRegion2AMI", {
-            "eu-west-1": {"AMI": "ami-00d88f77"},
-        })
+        if 'os' in self.data['ec2'] and self.data['ec2']['os'] == 'windows2012':
```
sevenmachines, line 94: If we're adding support to choosing the AMI, we might want to make this a bit more expansive. i.e
if 'os' in self.data['ec2']:
  if self.data['ec2']['os'] == 'windows2012':
    pass
  if self.data['ec2']['os'] == 'ubuntu14.04':
    pass
  else:
   default to ubuntu 14.04 or something?

sevenmachines, line 94: Not a biggie though

setup.py
```diff
@@ -4,7 +4,7 @@
 
 setup(
     name='bootstrap_cfn',
-    version='0.7.6',
+    version='0.7.7',
```
sevenmachines, line 7: we'll bump this as part of a release rather than a commit or PR

PR #184, opened 2016-02-19
Add cycle instances task

sevenmachines:
THis change adds a task to cycle the instances on an autoscaling
group, plus additional supporting methods for the Autoscale
class. This functionality means that we can recreate instances
sequentially in an autoscaling group, useful when we want to ensure
we are using the latest AMI's with the newest security fixes.

### DIFF
--- a/bootstrap_cfn/autoscale.py
+++ b/bootstrap_cfn/autoscale.py
@@ -1,6 +1,14 @@
+import logging
+
+import time
+
 import boto.ec2.autoscale
 
+import boto3
+
 from bootstrap_cfn import utils
+
+from bootstrap_cfn.errors import AutoscalingGroupNotFound, AutoscalingInstanceCountError
 
 
 class Autoscale:
@@ -27,7 +35,8 @@ class Autoscale:
                 resource_id=self.group.name)
 
             self.conn_asg.create_or_update_tags([tag])
-            print "Created ASG Tag: Tag({0}, {1})".format(key, value)
+            logging.getLogger("bootstrap-cfn").info("Created ASG Tag: Tag({0}, {1})"
+                                                    .format(key, value))
             return True
 
     def get_all_autoscaling_groups(self):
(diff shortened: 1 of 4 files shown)


sevenmachines:
@bennythejudge WIP task on cycling instances, seems to work ok on testing. If you fancy using this as the base of our security task I think it might be the way to go.

sevenmachines:
We need to add a test to this to make sure that at least one healthcheck is run on the instance before we pass it as healthy. The ASG set the new instance to healthy while its in its 'grace' period and only checks the health after that time has expired. Solutions could be,
- (best) Check that the ASG has run a healthcheck on the new instance before judging it 'healthy'
- (works-but-not-great) Wait until after the grace period before checking instance health

bootstrap_cfn/autoscale.py
```diff
@@ -46,3 +53,145 @@ def get_all_autoscaling_groups(self):
...
+        # is available. In that case we want to delay termination for a long enough time to hope the service
+        # sets itself up.
+        if not termination_delay and self.group.health_check_type != "ELB":
+            termination_delay = 360
+
+        # Get a list of the current instances
+        current_instance_ids = [instance.get('InstanceId') for instance in self.get_healthy_instances()]
+        logging.getLogger("bootstrap-cfn").info("cycle_instances: Found {} instance ids, {}"
+                                                .format(len(current_instance_ids), current_instance_ids))
+
```
filipposc5, line 82: Is it worth presenting a status of all instances at this stage (not just the healthy ones)? Assuming someone will be changing their states it might be worth giving out a quick check like: 
`Healthy: 3  Unhealthy: 0`  <-- this would at least make you feel better that there were 0 unhealthy
Or something like `Healthy:  i-whatever1, i-whatever2    Other: i-whatever3`  <-- even better, it shows you which was not healthy/terminating/changing to begin with.

This would help us track "what happened" in case for example one of the instances was not healthy to begin with.

bootstrap_cfn/autoscale.py
```diff
@@ -46,3 +53,145 @@ def get_all_autoscaling_groups(self):
...
+        new_instance_ids = [instance.get('InstanceId') for instance in self.get_healthy_instances()]
+        logging.getLogger("bootstrap-cfn").info("cycle_instances: {} instances recycled, {}"
+                                                .format(len(current_instance_ids), current_instance_ids))
+        logging.getLogger("bootstrap-cfn").info("cycle_instances: {} instances created, {}"
+                                                .format(len(new_instance_ids), new_instance_ids))
+
+    def set_autoscaling_desired_capacity(self,
+                                         capacity):
+        """
+        Set the desired instances count on an autosaling group
```
filipposc5: "autoscaling"

sevenmachines:
note: needs squishing

sevenmachines:
Did final squish

PR #187, opened 2016-02-24
Add a Name tag to instances

sevenmachines:
Add a Name tag to instances

The AWS console uses the Name tag on instances in its display. This
change sets this tag to the stack name to help identification when
viewing instances through the AWS console. The default format
for the resource Name key is,
`<application>-<environment>-<stack_hash>-<resource_type>`

### DIFF
--- a/bootstrap_cfn/config.py
+++ b/bootstrap_cfn/config.py
@@ -1021,6 +1021,9 @@ class ConfigParser(object):
         # Allow deprecation of tags
         ec2_tags = []
         deprecated_tags = ["Env"]
+        # Add a name tag for easy ec2 instance identification in the AWS console
+        ec2_tags.append(self._get_default_resource_name_tag(type="ec2"))
+        # Get all tags from the config
         for k, v in data['tags'].items():
             if k not in deprecated_tags:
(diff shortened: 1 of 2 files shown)


ltsampros:
I'd be happy to merge a change like this but current tests tend to disagree.

mattpep:
Whilst the patch is being developed, I'd also like additional resources to be tagged: security groups, VPCs, etc. There's probably no need to be exhaustive in this commit (future resource types later discovered to be missing can be added later) but there's merit in capturing the most common types of resource we create in a stack.

sevenmachines:
Don't listen to the tests :)Will fix tomorrow. Think it might be worth genericising the change to encompass other resources _get_default_resource_name or similar. Maybe with the type of resource included? Will have chat tomorrow.

sevenmachines:
@ltsampros @mattpep made it a little more generic with a 
 `<application>-<environment>-<stack_hash>-<resource_type>` format,
for example,
`helloniall-dev-ed4afa41-ec2`

Other resources would obviously end up as 
`helloniall-dev-ed4afa41-vpc` or whatever

PR #191, opened 2016-03-01
Override AMI

mattpep:
- Provide the ability to override the AMI used
- Use the AWS Windows 2012 R2 AMI instead of RTM


ltsampros:
LGTM.

PR #194, opened 2016-03-07
[WIP] Allows to override the aws profile for route53

filipposc5:
In certain cases the aws stack belongs to a different sub account
than the route53 domains so certain operations could benefit by
crossing to a different sub account, using different credentials.

For now only the enter_maintenance and exit_maintenance tasks are
aware of this change.


filipposc5:
This isn't needed anymore.

PR #201, opened 2016-04-19
Setup boto3 default session with default region

filipposc5:
Reason this is useful is because the workaround would be to specify for each of the aws profiles a separate entry in `~/.aws/config` in the format of: 

```
[profile profile_name]
aws_region = eu-west-1
```

With this change it will pick up the default region from "env", which is setdefault to "eu-west-1"


ltsampros:
Looks good.

PR #202, opened 2016-04-21
add dry_run option to the maintenance tasks

filipposc5:
Usage:
- `enter_maintenance:1.1.1.1,dry_run=True`
- `exit_maintenance:dry_run=True`


chriswood123:
I think this is definitely a good idea, would make running this command safer and less scary.

PR #209, opened 2016-06-15
Use a configuration file for default cloudformations settings

sevenmachines:
Currently we have some default settings coded into config.py.
This change introduces the use of a configuration file to
specify default settings for cloudformation. These settings are
replaced by those placed into the project configuration files.

In particular, this means that we can provide defaults for each
environments seperately, putting less resources on dev environments
than on prod. In this case we need substantial RDS instance types
in prod to support defaulting to encryption. This is replicated
across all environments. With this change we can set the dev
environment to use no encryption and small instances as default.


sevenmachines:
As a note, theres the beginning of some basic validation calls, i've kept these minimal for the moment with a view to perhaps expanding them out into another 'lint' type class with some deeper validation

PR #211, opened 2016-06-16
Add random number generator for RDS credentials

sevenmachines:
This change removes the requirement for RDS credentials and instead
allows them to be generated using a simple random string generator.
Combined with setting resource defaults, it means that RDS instances
can be sensibly generated from only an empty key,
rds: {}

### DIFF
--- a/bootstrap_cfn/config.py
+++ b/bootstrap_cfn/config.py
@@ -481,12 +481,12 @@ class ConfigParser(object):
         """
         # REQUIRED FIELDS MAPPING
         required_fields = {
-            'db-name': 'DBName',
-            'db-master-username': 'MasterUsername',
-            'db-master-password': 'MasterUserPassword',
         }
 
         optional_fields = {
+            'db-name': 'DBName',
+            'db-master-username': 'MasterUsername',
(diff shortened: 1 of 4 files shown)


ltsampros:
@niallcreech why are we turning the default rds encryption to false in the code again?

sevenmachines:
@ltsampros you thinking of the other PR?
The defaults change is...
dev environments default to un-encrypted
all other default to encrypted

prod and staging (and anything not dev) are still always encrypted, the place its done is just move to the defaults file rather than in code.

[EDIT] Note the price differential in moving from db.t2.small to the large required by encryption, and applying that across all dev instances. When by definition they should not ever have any sensitive information on them

ltsampros:
@niallcreech Ooops. Sorry I was mainly talking about the other PR.

sevenmachines:
We overwrite our cloudformation.sls pillar data everytime we run a salt update, as such, we'd lose any usernames and passwords we set randomly in there. Probably the easiest solution is to not overwrite the cloudformation after stack creation as it hasn't actually changed on the stack. But this moves a trivial change to a more substantial change so going to bin this for now.

PR #215, opened 2016-07-04
Core multiple stacks management

yufangzhang:
- Modify fab task cfn_create, cfn_delete to mange multiple stacks.
- Create task set_active_stack(tag) to switch among stacks
- Add unit tests for new methods

### DIFF
--- a/bootstrap_cfn/config.py
+++ b/bootstrap_cfn/config.py
@@ -110,7 +110,7 @@ class ConfigParser(object):
     def __init__(self, data, stack_name, environment=None, application=None):
         self.stack_name = stack_name
         self.data = data
-
+        self.stack_id = self.stack_name.split('-')[-1]
         # Some things possibly used in user data templates
         self.environment = environment
         self.application = application
@@ -551,6 +551,12 @@ class ConfigParser(object):
         if 'db-engine' in self.data['rds'] and self.data['rds']['db-engine'].startswith("sqlserver"):
             required_fields.pop('db-name')
 
+        if 'identifier' in self.data['rds']:
+            # update identifier name
+            self.data['rds']['identifier'] = "{}-{}".format(self.data['rds']['identifier'], self.stack_id)
+            logging.info("identifier was updated to {}".format(self.data['rds']['identifier']))
+            print "identifier was updated to {}".format(self.data['rds']['identifier'])
+
         # TEST FOR REQUIRED FIELDS AND EXIT IF MISSING ANY
         for yaml_key, rds_prop in required_fields.iteritems():
             if yaml_key not in self.data['rds']:
@@ -813,7 +819,7 @@ class ConfigParser(object):
                 RecordSets=[
                     RecordSet(
                         "TitleIsIgnoredForThisResource",
-                        Name="%s.%s" % (elb['name'], elb['hosted_zone']),
+                        Name="%s-%s.%s" % (elb['name'], self.stack_id, elb['hosted_zone']),
                         Type="A",
                         AliasTarget=AliasTarget(
                             GetAtt(load_balancer, "CanonicalHostedZoneNameID"),
--- a/bootstrap_cfn/errors.py
+++ b/bootstrap_cfn/errors.py
@@ -77,3 +77,39 @@ class AutoscalingInstanceCountError(BootstrapCfnError):
             "Could not find {} instances in autoscaling group {}. Actual state is {} instances, {}"
             .format(expected_instance_count, autoscaling_group, len(instances), instances)
         )
+
+
+class TagRecordExistConflictError(BootstrapCfnError):
+    def __init__(self, stack_tag):
+        msg = ("An {} record already exists. ".format(stack_tag))
+        super(TagRecordExistConflictError, self).__init__(msg)
+
+
+class ActiveTagExistConflictError(BootstrapCfnError):
+    def __init__(self, stack_id):
+        msg = ("An active record already exists in {}. ".format(stack_id))
+        super(ActiveTagExistConflictError, self).__init__(msg)
+
+
+class TagRecordNotFoundError(BootstrapCfnError):
+    def __init__(self, tag):
+        msg = ("Could not find a dns record for tag '{}'. ".format(tag))
+        super(TagRecordNotFoundError, self).__init__(msg)
+
+
+class PublicELBNotFoundError(BootstrapCfnError):
+    def __init__(self):
+        msg = "Could not find an internet facing ELB according to cloudformation configuration. "
+        super(PublicELBNotFoundError, self).__init__(msg)
+
+
+class StackRecordNotFoundError(BootstrapCfnError):
+    def __init__(self, stack_record_name):
+        msg = ("Could not find a dns record for stack '{}'. ".format(stack_record_name))
+        super(StackRecordNotFoundError, self).__init__(msg)
+
+
+class UpdateDNSRecordError(BootstrapCfnError):
+    def __init__(self):
+        msg = "Error updating dns record. "
+        super(UpdateDNSRecordError, self).__init__(msg)
--- a/bootstrap_cfn/fab_tasks.py
+++ b/bootstrap_cfn/fab_tasks.py
@@ -7,8 +7,6 @@ import uuid
 
 import boto3
 
-import dns.resolver
-
 from fabric.api import env, task
 from fabric.colors import green, red
 from fabric.utils import abort
@@ -17,10 +15,10 @@ from bootstrap_cfn.autoscale import Autoscale
 from bootstrap_cfn.cloudformation import Cloudformation
 from bootstrap_cfn.config import ConfigParser, ProjectConfig
 from bootstrap_cfn.elb import ELB
-from bootstrap_cfn.errors import ActiveTagExistConflictError, BootstrapCfnError,\
-    CfnConfigError, CloudResourceNotFoundError, DNSRecordNotFoundError,\
-    PublicELBNotFoundError, StackRecordNotFoundError, TagRecordExistConflictError,\
-    TagRecordNotFoundError, UpdateDNSRecordError, ZoneIDNotFoundError
+from bootstrap_cfn.errors import (ActiveTagExistConflictError, BootstrapCfnError,
+                                  CfnConfigError, CloudResourceNotFoundError, DNSRecordNotFoundError,
+                                  PublicELBNotFoundError, StackRecordNotFoundError, TagRecordExistConflictError,
+                                  TagRecordNotFoundError, UpdateDNSRecordError, ZoneIDNotFoundError)
 from bootstrap_cfn.iam import IAM
 from bootstrap_cfn.r53 import R53
 from bootstrap_cfn.utils import tail
@@ -351,9 +349,9 @@ def get_stack_name(new=False):
         # For back-compatibility
         set_stack_name()
 
-    if hasattr(env, 'tag'):
-        stack_tag = env.tag
-    else:
+    try:
+        stack_tag = get_env_tag()
+    except AttributeError:
         stack_tag = 'active'
         env.tag = stack_tag
     if not hasattr(env, 'stack_name'):
@@ -361,21 +359,21 @@ def get_stack_name(new=False):
         # know it yet...
         env.stack_name = 'temp'
         zone_name = get_zone_name()
-        if not zone_name:
-            raise CfnConfigError("No master_zone in yaml, unable to create/find DNS records for stack name")
-        logger.info("fab_tasks::get_stack_name: Found master zone '{}' in config...".format(zone_name))
+        zone_id = get_zone_id()
+        logger.info("fab_tasks::get_stack_name: Found master zone '%s' in config...", zone_name)
         # get record name in the format of: stack.[stack_tag].[app]-[env]
         record_name = get_tag_record_name(stack_tag)
         dns_name = "{}.{}".format(record_name, zone_name)
+        r53_conn = get_connection(R53)
         try:
             # get stack id
-            stack_suffix = dns.resolver.query(dns_name, 'TXT')[0].to_text().replace('"', "")
-            logger.info("fab_tasks::get_stack_name: Found stack suffix '{}' "
-                        "for dns record '{}'... ".format(stack_suffix, dns_name))
+            stack_suffix = r53_conn.get_record(zone_name, zone_id, record_name, 'TXT').replace('"', "")
+            logger.info("fab_tasks::get_stack_name: Found stack suffix '%s' "
+                        "for dns record '%s'... ", stack_suffix, dns_name)
             legacy_name = get_legacy_name()
             env.stack_name = "{0}-{1}".format(legacy_name, stack_suffix)
-            logger.info("fab_tasks::get_stack_name: Found stack name '{}'...".format(env.stack_name))
-        except dns.resolver.NXDOMAIN:
+            logger.info("fab_tasks::get_stack_name: Found stack name '%s'...", env.stack_name)
+        except Exception:
             raise DNSRecordNotFoundError(dns_name)
 
     return env.stack_name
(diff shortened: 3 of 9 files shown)


bootstrap_cfn/fab_tasks.py
```diff
@@ -393,39 +397,90 @@ def set_stack_name():
...
+    Check if stack_tag is in use
+    Args:
+        stack_tag: the tag of stack
+    Returns:
+        True if stack exists
+
+    """
+    r53_conn = get_connection(R53)
+    zone_id = get_zone_id()
+    record = "{}".format(get_tag_record_name(stack_tag))
```
mattpep: This could probably be simply:

``` python
record = get_tag_record_name(stack_tag)
```

i.e. without the call to `format`.

bootstrap_cfn/errors.py
```diff
@@ -77,3 +77,33 @@ def __init__(self, autoscaling_group, expected_instance_count, instances):
...
+
+class TagRecordExistConflictError(BootstrapCfnError):
+    def __init__(self, stack_tag):
+        msg = ("An {} record already exists. ".format(stack_tag))
+        super(TagRecordExistConflictError, self).__init__(msg)
+
+
+class ActiveTagExistConflictError(BootstrapCfnError):
+    def __init__(self, stack_id):
+        msg = ("An active record already exists in. ".format(stack_id))
```
mattpep: The `format` parameter is not used. Presumably something like the following is intended:

``` python
msg = 'An active record already exists in {}.'.format(stack_id)
```

bootstrap_cfn/r53.py
```diff
@@ -68,14 +69,71 @@ def update_dns_record(self, zone, record, record_type, record_value, is_alias=Fa
...
+    def get_record(self, zone_name, zone_id, record_name, record_type):
+        """
+
+        Args:
+            zone_name:
+            zone_id:
+            record_name(String):
+            record_type:
+        Returns:
+            String
```
mattpep: I suggest adding the following text to this line:

`or None, in the event of there being no A or TXT record`

bootstrap_cfn/fab_tasks.py
```diff
@@ -701,3 +802,118 @@ def cycle_instances(delay=None):
...
+        print green("Active stack id is: {}".format(active_stack_id))
+        return active_stack_id
+    else:
+        print green("No active stack exists.")
+        return None
+
+
+def get_all_elbs():
+    """
+    Returns all available elbs from cloudformation configuration
```
mattpep: The docstring should be a bit more clear in its description -e.g.

`Returns all internet-facing elbs from cloudformation configuration`

sevenmachines:
@yufangzhang happy to merge after those couple of minor things @mattpep mentioned are fixedup!

yufangzhang:
@mattpep I've done the changes. can you have a look ~?

sevenmachines:
@mattpep @yufangzhang we happy to merge?

mattpep:
LGTM. Ideally we should squash the fixup commit into the one it updates. Then it's good to merge.

bootstrap_cfn/fab_tasks.py
```diff
@@ -393,39 +397,90 @@ def set_stack_name():
...
+    return env.stack_name
+
+
+def hastag(stack_tag):
+    """
+    Check if stack_tag is in use
+    Args:
+        stack_tag: the tag of stack
+    Returns:
+        True if stack exists
```
filipposc5: Is this accurate? To be specific does it return "True" if stack exists?

bootstrap_cfn/fab_tasks.py
```diff
@@ -483,35 +540,80 @@ def cfn_delete(force=False, pre_delete_callbacks=None):
...
-
-    # Wait for stacks to delete
-    print 'Waiting for stack to delete.'
-
-    tail(cfn, stack_name)
+    r53_conn = get_connection(R53)
+    if not force:
+        x = raw_input("Are you really sure you want to blow away the whole stack for {}!? (y/n)\n".format(stack_name))
+        if x not in ['y', 'Y', 'Yes', 'yes']:
+            sys.exit(1)
```
filipposc5: Why did you move this?

bootstrap_cfn/fab_tasks.py
```diff
@@ -483,35 +540,80 @@ def cfn_delete(force=False, pre_delete_callbacks=None):
...
+        if stack_record_object and stack_record_object.to_print() == main_record_object.to_print():
+            r53_conn.delete_dns_record(get_zone_id(), main_record_name, 'A', main_record_value, is_alias=True)
+
+        # delete 'TXT' record
+        tag_record_name = get_tag_record_name(stack_tag)
+        record_value = '"{}"'.format(r53_conn.get_record(
+            get_zone_name(), get_zone_id(), tag_record_name, 'TXT'))
+        record_name = '{}.{}'.format(tag_record_name, get_zone_name())
+        if stack_id and stack_id == record_value[1:-1]:
+            r53_conn.delete_dns_record(get_zone_id(), record_name, 'TXT', record_value)
```
filipposc5: In this method you seem to be using get_zone_id() about 9 times. Does this get cached by the boto library? We shouldn't be repeating calls pointlessly as there is a risk to get throttled. 
Also I had created `get_cached_zone_id` that will do some caching when it looks up zone ids. Actually the whole get_zone_id() can be replicated by refactoring to this and passing an empty cache dict if you don't want caching: 

``` python
def get_zone_id(r53_conn):
    return get_cached_zone_id(r53_conn, {}, get_zone_name())
```

bootstrap_cfn/fab_tasks.py
```diff
@@ -701,3 +802,118 @@ def cycle_instances(delay=None):
...
+        if x not in ['y', 'Y', 'Yes', 'yes']:
+            sys.exit(1)
+
+    # update TXT record
+    try:
+        r53_conn.update_dns_record(zone_id, "{}.{}".format(active_record, get_zone_name()), 'TXT',
+                                   '"{}"'.format(tag_stack_id))
+        logger.info("fab_tasks::set_active_stack: Successfully updated dns alias record")
+    except Exception as e:
+        logger.error("fab_tasks::set_stack_name:"
```
filipposc5: `set_active_stack` instead of `set_stack_name`

bootstrap_cfn/fab_tasks.py
```diff
@@ -701,3 +802,118 @@ def cycle_instances(delay=None):
...
+                    record_object.alias_dns_name,
+                    record_object.alias_evaluate_target_health]
+    if record_value:
+        # point [helloworld.dsd.io] to [helloworld-12345.dsd.io]'s ELB
+        try:
+            r53_conn.update_dns_record(zone_id, main_record_name, 'A', record_value, is_alias=True)
+            logger.info("fab_tasks::set_active_stack: Successfully updated dns alias record")
+            print green("Active stack is switched to {}".format(tag_record))
+        except Exception as e:
+            logger.error("fab_tasks::set_stack_name:"
```
filipposc5: `set_active_stack` instead of `set_stack_name`

bootstrap_cfn/fab_tasks.py
```diff
@@ -701,3 +802,118 @@ def cycle_instances(delay=None):
...
+    # update TXT record
+    try:
+        r53_conn.update_dns_record(zone_id, "{}.{}".format(active_record, get_zone_name()), 'TXT',
+                                   '"{}"'.format(tag_stack_id))
+        logger.info("fab_tasks::set_active_stack: Successfully updated dns alias record")
+    except Exception as e:
+        logger.error("fab_tasks::set_stack_name:"
+                     "Error updating dns record: "
+                     "{}".format(e.message))
+
```
filipposc5: Is there supposed to be a `raise` here? If the record fails to get updated?

bootstrap_cfn/fab_tasks.py
```diff
@@ -701,3 +802,118 @@ def cycle_instances(delay=None):
...
+    # get all public facing elbs
+    elb = get_one_public_elbs()
```
filipposc5: comment misleading. you are only getting the first.

bootstrap_cfn/fab_tasks.py
```diff
@@ -701,3 +802,118 @@ def cycle_instances(delay=None):
...
+def get_all_elbs():
+    """
+    Returns all internet-facing elbs from cloudformation configuration
+    """
+    cfn_config = get_config()
+    elbs = [x.get('name') for x in cfn_config.data.get('elb', {}) if x.get('scheme') == 'internet-facing']
+    return elbs
+
+
+def get_one_public_elbs():
```
filipposc5: consider renaming to `get_first_public_elb` or something more appropriate.

bootstrap_cfn/fab_tasks.py
```diff
@@ -701,3 +802,118 @@ def cycle_instances(delay=None):
...
+    """
+    Returns the first elb if exists.
+    """
+    elbs = get_all_elbs()
+    if len(elbs) < 1:
+        raise PublicELBNotFoundError
```
filipposc5, line 871: Is this really an invalid case, not having a public ELB?

bootstrap_cfn/config.py
```diff
@@ -51,7 +51,7 @@ class ConfigParser(object):
     def __init__(self, data, stack_name, environment=None, application=None):
         self.stack_name = stack_name
         self.data = data
-
+        self.stack_id = self.stack_name.split('-')[-1]
```
filipposc5, line 113: how can you have access to the stack id here? since this generates the template to create the stack id to begin with.  You might be able to use  `AWS::StackId` (or maybe `AWS::StackName` ) with cloudformation syntax to obtain the ID.

bootstrap_cfn/fab_tasks.py
yufangzhang: because when it comes to "deleting active stack", I want to delete "active" dns records while keep the stack. 
The reason I put cfn.delete after dns deleted is though, previously it stucked in vpc deleting. I wanna put it at the end so that I can killed the program and delete vpc manually in AWS console if neccessary.

bootstrap_cfn/fab_tasks.py
yufangzhang: actually they're the same id...I'll remove the duplicated calls.

bootstrap_cfn/fab_tasks.py
yufangzhang: a yeah

bootstrap_cfn/fab_tasks.py
yufangzhang, line 871: currently we find internet facing elbs from cloudformation so if we don't get any, should check its configuration file...I'll specify it in the error message of PublicELBNotFoundError

bootstrap_cfn/fab_tasks.py
mattpep: and if there is more than one public ELB, how are they ordered? In other words, how can one know which will be the first, and whether it will be the correct one?

yufangzhang:
@mattpep yeup, updated!

bootstrap_cfn/config.py
yufangzhang, line 113: ah we created stack_name in the format of [app-env-stackid] so can access it...

mattpep:
The first bullet point of this PR makes reference to changing the task `cfn_create` but I see no changes of that task within `config.py`.
Does the stack_id now have to be specified? More generically, it'd be useful for the documentation be updated to explain how these new tasks work.

bootstrap_cfn/config.py
filipposc5, line 113: Disregard this comment of mine, as we saw this is fine.

bootstrap_cfn/fab_tasks.py
filipposc5, line 871: All I am saying is that there are two cases where we can create a stacks that don't fit this restriction you have imposed. 1) A stack without an ELB 2) a stack with a non internet facing ELB. I haven't tested it but at least the 2nd case is valid I believe. So maybe raising an exception is not suitable here.

yufangzhang:
@mattpep we generate `stack_id` from `uuid` in `set_stack_name` at the beginning as what it used to be, then add it to `stack_name`. In config.py, we only added `self.stack_id` as instance variable in config and changed the format of dns record name.

bootstrap_cfn/r53.py
```diff
@@ -68,14 +69,71 @@ def update_dns_record(self, zone, record, record_type, record_value, is_alias=Fa
...
+        """
+        record_fqdn = "{0}.{1}.".format(record_name, zone_name)
+        rrsets = self.conn_r53.get_all_rrsets(zone_id, type=record_type, name=record_fqdn)
         for rr in rrsets:
-            if rr.type == record_type and rr.name == fqdn:
+            if rr.type == record_type and rr.name == record_fqdn:
                 if rr.type == 'TXT':
                     rr.resource_records[0] = rr.resource_records[0][1:-1]
+                if rr.type == 'A':
+                    return rr.alias_dns_name
```
filipposc5: An A record doesn't have to be a DNS alias.

yufangzhang: yes. But we wanted to keep returns as String type instead of "either String( `TXT` ) or RecordObejct ( `A` )".

bootstrap_cfn/fab_tasks.py
```diff
@@ -393,39 +396,70 @@ def set_stack_name():
...
         env.tag = stack_tag
-    legacy_name = "{0}-{1}".format(env.application, env.environment)
-    # get_config needs a stack_name so this is a hack because we don't
-    # know it yet...
-    env.stack_name = 'temp'
-    cfn_config = get_config()
-    r53_conn = get_connection(R53)
-    zone_name = cfn_config.data.get('master_zone', None)
+    record = "{}.{}".format(get_tag_record_name(stack_tag), zone_name)
+    logger.info("fab_tasks::set_stack_name: "
```
cliffxuan, line 413: `format` is better than using `%s` in formatting a string, however, in `logging`, `%s` is actually preferred, so you would do:

```
logger.info("Creating stack suffix '%s' for record '%s'... %s...", stack_suffix, record, zone_id)
```

bootstrap_cfn/fab_tasks.py
```diff
@@ -393,39 +396,70 @@ def set_stack_name():
     dns records to retreive it in the future.
 
     """
+    # create a stack id
+    r53_conn = get_connection(R53)
+    zone_name = get_zone_name()
+    zone_id = get_zone_id()
+    stack_suffix = uuid.uuid4().__str__()[-8:]
```
cliffxuan, line 401: you can do `uuid.uuid4().hex[-8:]`

bootstrap_cfn/fab_tasks.py
```diff
@@ -393,39 +396,70 @@ def set_stack_name():
...
     if hasattr(env, 'tag'):
```
cliffxuan: EAFP is a more pythonic way

```
try:
      if env.tag == 'active':
      ....
except AttributeError:
    env.tag = stack_suffix 
```

bootstrap_cfn/fab_tasks.py
```diff
@@ -393,39 +396,70 @@ def set_stack_name():
...
+    record = "{}.{}".format(get_tag_record_name(stack_tag), zone_name)
+    logger.info("fab_tasks::set_stack_name: "
+                "Creating stack suffix {} "
+                "for record '{}' "
+                "in zone id '{}'...".format(stack_suffix, record, zone_id))
+    # Let DNS update DNSServerError propogate
+    try:
+        r53_conn.update_dns_record(zone_id, record, 'TXT', '"{0}"'.format(stack_suffix))
+        env.stack_name = "{0}-{1}".format(get_legacy_name(), stack_suffix)
+    except:
```
cliffxuan: always use `except Exception:` instead of just `except:` to trap exceptions in general. `except` traps other things like `KeyboardInterrupt`.

bootstrap_cfn/fab_tasks.py
```diff
@@ -393,39 +396,70 @@ def set_stack_name():
...
+        raise UpdateDNSRecordError
+    return env.stack_name
+
+
+def get_zone_name():
+    zone_name = get_basic_config().get('master_zone', None)
```
cliffxuan: better do the EAFP way:

```
try:
    zone_name = get_basic_config()['master_zone']
except KeyError:
    raise  CfnConfigError("No master_zone in yaml, unable to create/find DNS records for stack name")
```

bootstrap_cfn/fab_tasks.py
```diff
@@ -701,3 +763,114 @@ def cycle_instances(delay=None):
...
+        zone_name = get_zone_name()
+        zone_id = get_zone_id()
+        active_stack_id = r53_conn.get_record(zone_name, zone_id, active_record, 'TXT')
+        elb = get_first_public_elb()
+        dns_record_name = '{}-{}'.format(elb, active_stack_id)
+        dns_record_value = r53_conn.get_record(zone_name, zone_id, dns_record_name, 'A')
+        main_record_value = r53_conn.get_record(zone_name, zone_id, elb, 'A')
+    except:
+        print green("No active stack exists.")
+        return None
```
cliffxuan: could just do `return`, which is the same as `return None`

bootstrap_cfn/fab_tasks.py
```diff
@@ -18,7 +15,10 @@
 from bootstrap_cfn.cloudformation import Cloudformation
 from bootstrap_cfn.config import ConfigParser, ProjectConfig
 from bootstrap_cfn.elb import ELB
-from bootstrap_cfn.errors import BootstrapCfnError, CfnConfigError, CloudResourceNotFoundError, DNSRecordNotFoundError, ZoneIDNotFoundError
+from bootstrap_cfn.errors import ActiveTagExistConflictError, BootstrapCfnError,\
```
cliffxuan: preferred way to write long import line is to use `()` rather than `\` https://www.python.org/dev/peps/pep-0328/

```
from bootstrap_cfn.errors import (ActiveTagExistConflictError, BootstrapCfnError,
    CfnConfigError, CloudResourceNotFoundError, DNSRecordNotFoundError,
    ...)
```

yufangzhang:
@cliffxuan Thanks for the advice Cliff! I've made some changes to make it more pythonic 👻 
@lukaszraczylo think you may have used this to do the CCCD stuff? I've just updated a bug on deleting active stack records. hopefully you haven't come across that....
@filipposc5 @niallcreech I think this beast is ready to be merged, can anyone have a look? sorry about its size :P

PR #218, opened 2016-07-15
(WIP) Stack Version Tag initial checkin. 

ezman:
Needs testing this branch exists so builds can be tested.


bootstrap_cfn/fab_tasks.py
```diff
@@ -622,7 +621,8 @@ def get_cloudformation_tags():
     """
     return {
         "Env": env.environment,
-        "Application": env.application
+        "Application": env.application,
+        "Bootstrap-cfn Version": bootstrap_cfn.__version__
```
mattpep, line 692: Since this is a hash, the order of the key-value pairs don't matter. It'd be neater to put the `Bootstrap-cfn Version` tag before `Application` such that only 1 line appears in the diff output.

lukaszraczylo:
@ezman conflicts need fixing + it should pass travis tests as well :)

setup.py
```diff
@@ -1,10 +1,11 @@
 #!/usr/bin/env python
 
 from setuptools import find_packages, setup
+import pkg_resources
 
 setup(
     name='bootstrap_cfn',
-    version='0.11.2',
+    version=pkg_resources.require("bootstrap-cfn")[0].version,
```
sevenmachines, line 8: Not sure where we're setting the version string? options are, probably easiest to make setup.py the source of truth on the version and just version='x.x.x' here?

bootstrap_cfn/fab_tasks.py
sevenmachines: you'll probably need an import bootstrap_cfn somewhere, or use pkg_resources.require("bootstrap-cfn")[0].version directly

sevenmachines:
@ezman see comments. But also, need to rebase to master and squash commits now we add versioneer, and remove it in a later commit. It's a small change so one commit should cover it all

sevenmachines:
covered by https://github.com/ministryofjustice/bootstrap-cfn/pull/232

PR #223, opened 2016-08-18
Fixup get_stack_list

yufangzhang:
Add method for get_stack_list to return all resource records


ezman:
I have merged the PR - but we need a test case for the jenkins integration test.

PR #226, opened 2016-08-25
Add KeyName to fab parameters

yufangzhang:
This commit adds keyname as a fab parameter.
with it, you can set it to your keypair in AWS
KeyName defined in config file will not be used.

### DIFF
--- a/bootstrap_cfn/fab_tasks.py
+++ b/bootstrap_cfn/fab_tasks.py
@@ -35,6 +35,7 @@ env.setdefault('config')
 env.setdefault('stack_passwords')
 env.setdefault('blocking', True)
 env.setdefault('aws_region', 'eu-west-1')
+env.setdefault('keyname', None)
 
 # GLOBAL VARIABLES
 TIMEOUT = 3600
@@ -183,6 +184,18 @@ def user(username):
         variable to.
     """
     env.user = username
+
+
+@task
+def keyname(keyname):
+    """
+    Sets the keyname to the keypair name on AWS
+
+    Sets the keyname to specific keypair name you created instead of "default"
+    Args:
+        keyname: the name of keypair on AWS
+    """
+    env.keyname = str(keyname).lower()
 
 
 @task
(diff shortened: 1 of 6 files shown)


bootstrap_cfn/fab_tasks.py
```diff
@@ -478,6 +491,13 @@ def _validate_fabric_env():
     elif not os.path.isfile(env.config):
         sys.exit("\n[ERROR] Config file %s does not exist" % str(env.config))
 
+    # keyname is mandatory in cfn_create, optional in others.
+    # if _validate_fabric_env() is called by cfn_create, then check keyname.
+    # otherwise not.
+    if called_by_cfn_create:
+        if env.keyname is None:
```
ltsampros: Could you please put them on the same line?

bootstrap_cfn/fab_tasks.py
```diff
@@ -607,6 +628,8 @@ def cfn_create(test=False):
     specification will be generated and used to create a
     stack on AWS.
     """
+    _validate_fabric_env(True)
```
ltsampros: Please add a keyword argument here to improve readability and show intent.

bootstrap_cfn/fab_tasks.py
yufangzhang: ahhh sure

bootstrap_cfn/fab_tasks.py
```diff
@@ -478,6 +491,12 @@ def _validate_fabric_env():
     elif not os.path.isfile(env.config):
         sys.exit("\n[ERROR] Config file %s does not exist" % str(env.config))
 
+    # keyname is mandatory in cfn_create, optional in others.
+    # if _validate_fabric_env() is called by cfn_create, then check keyname.
+    # otherwise not.
+    if called_by_cfn_create and env.keyname is None:
+            sys.exit("\n[ERROR] Please specify a keyname, e.g 'keyname:keyops'")
+
```
sevenmachines: I can appreciate why it's a good idea to mandate it, but, this is a breaking change. A reasonable compromise might be to accept that some stacks have it defined in their config, and warn rather than error. Also, deleting the keyname from the template-deploy example?

yufangzhang: do you mean it's better to accept the param in both config file and command line, or just warn if it's defined in config rather than fab params?

sevenmachines: Maybe accept it if its in the config, but warning that it's deprecated/not-advised. whatever
Only a suggestion to think about though, it's not a requirement. If you're sure this is the best way then fire on

PR #235, opened 2016-12-06
Update default RDS version to 9.5.4

sevenmachines:
The default for RDS in AWS is now 9.5.4, this change matches that.


ltsampros:
Looks good to me.

PR #236, opened 2016-12-08
Markup fix

mattpep:


mattpep:
I'm undecided whether to squish everything into one. I'm slightly of the opinion of keeping the commits separate because there is a lot of change in this PR, but am open to opinions in either direction.

PR #238, opened 2016-12-14
get elb helpers shouldn't raise if none elbs exist.

ltsampros:
**get elb helpers shouldn't raise if none elbs exist.**


mattpep:
These two commits should probably be squished into one since they're both for the same change in functionality.

But the general principle is sound and LGTM.

PR #240, opened 2016-12-19
Add UK ami mappings

sevenmachines:
This change updates the available mapping to include eu-west-2
AMI's.

### DIFF
--- a/bootstrap_cfn/config.py
+++ b/bootstrap_cfn/config.py
@@ -4,6 +4,7 @@ import os
 import sys
 import textwrap
 
+from fabric.api import env
 from fabric.colors import green
 
 from troposphere import Base64, FindInMap, GetAZs, GetAtt, Join, Output, Ref, Tags, Template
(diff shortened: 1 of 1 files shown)


bootstrap_cfn/config.py
```diff
@@ -1197,25 +1198,49 @@ def _get_os_data(self):
...
+              'windows2012': {
+                  'name': 'windows2012',
+                  'ami': 'ami-8519a9f6',
+                  'region': region,
+                  'distribution': 'windows',
+                  'type': 'windows',
+                  'release': '2015.12.31'
+              }
+          }
+        else:
```
mattpep, line 1257: It looks like the behaviour as proposed here is to mandate a value for `AWS_REGION`. Would it be better to have the default behaviour be to use `eu-west-1` - as that will preserve forwards compatibility?

sevenmachines: Looking to centralise the region information in env.aws_region and synching the boto3.DEFAULT_SESSION to that too. but then defaulting the region to eu-west-1 as you mention for backwards compatibility.

sevenmachines: Not sure I've done that yet though, might be a different PR, or sitting on the laptop somewhere

mattpep: I've just found https://github.com/ministryofjustice/bootstrap-cfn/pull/241 which is the change you describe. (Linking to it here to keep the thread complete)

bootstrap_cfn/config.py
```diff
@@ -1197,25 +1198,49 @@ def _get_os_data(self):
             OSTypeNotFoundError: Raised when the OS in the config file is not
                 recognised
         """
+        region=env.aws_region
         os_default = 'ubuntu-1404'
-        available_types = {
-            'ubuntu-1404': {
+        if region == 'eu-west-2':
+          available_types = {
+              'ubuntu-1604': {
```
ltsampros, line 1205: Shouldn't we have the corresponding image available in eu-west-1?

PR #241, opened 2016-12-19
Add a region argument to the aws task

sevenmachines:
This change allows the region used by the AWS profile we set on the
command line to be specified.


ltsampros:
Happy to merge if build checks are fixed.

PR #243, opened 2016-12-21
Disable production settings on staging

sevenmachines:
We are now not requiring prod performance instances on staging,
so this change will reduce the defaults.


ltsampros:
LGTM.

Issue #249, opened 2017-02-15
set_active_stack ignores internal load balancers.


ltsampros:
Per subject, if internal load balancers are defined in cloudformation file they get ignored by set_active_stack.

yufangzhang:
fixed by https://github.com/ministryofjustice/bootstrap-cfn/pull/251

PR #250, opened 2017-02-15
set_active_stack operates on all elbs

ltsampros:
**set_active_stack operates on all elbs**
**Fixes issuee #249**


bootstrap_cfn/fab_tasks.py
```diff
@@ -896,7 +896,7 @@ def set_active_stack(stack_tag, force=False):
     except Exception:
         raise UpdateDNSRecordError
 
-    elbs = get_public_elbs()
+    elbs = get_all_elbs()
```
filipposc5, line 899: This makes no difference

filipposc5 (changes requested):
Currently this does not achieve any different behaviour, because the pre-existing method names are confusing and still filter for scheme internet-facing

PR #253, opened 2017-02-23
(WIP) Remove DNS tag in the event of rollback

### LINKED ISSUE #221: failed cfn_create with tag doesn't delete the R53 entries after rollback.
After a failed cfn_create with a tag:active2, I tried recreating the stack with the same tag:active2 and I got the following error:
`bootstrap_cfn.errors.TagRecordExistConflictError: An active2 record already exists.`

I think we should be removing the related R53 entries after the rollback.

The steps to reproduce this are
```fab tag:active2 ....... cfn_create``` that fails
```fab tag:active2 ....... cfn_create``` that fails because the failed cfn_create didn't delete the R53 entries for tag:active2.

### PR DESCRIPTION
**Pull request by mattpep**
Fixes #221

### DIFF
--- a/bootstrap_cfn/fab_tasks.py
+++ b/bootstrap_cfn/fab_tasks.py
@@ -697,7 +697,17 @@ def cfn_create(test=False):
     """
     _validate_fabric_env()
     stack_name = get_stack_name(new=True)
+    stack_id = stack_name.split('-')[-1]
     cfn_config = get_config(called_by_cfn_create=True)
+    try:
+        stack_tag = get_env_tag()
+    except AttributeError:
+        stack_tag = 'active'
+        env.tag = stack_tag
+    r53_conn = get_connection(R53)
+    zone_name = get_zone_name()
+    zone_id = get_zone_id()
+    txt_tag_record = get_tag_record_name(stack_tag)
 
     cfn = get_connection(Cloudformation)
     if test:
@@ -738,7 +748,14 @@ def cfn_create(test=False):
         # So delete the SSL cert that we uploaded
         if 'ssl' in cfn_config.data:
             iam.delete_ssl_certificate(cfn_config.ssl(), stack_name)
-        abort('Failed to create stack: {0}'.format(stack))
+        # delete the DNS tag
+        logger.info("Deleting DNS tag for '{stack} from '{zone}'...".format(stack=stack_id, zone=zone_name))
+        try:
+            print green("Removing tag stack from DNS...")
+            r53_conn.delete_txt_record(zone_name, zone_id, txt_tag_record)
+        except boto.route53.exception.DNSServerError:
+            print red("Failed to remove DNS tag")
+        abort(red('Failed to create stack: {0}. You may want to delete it manually.'.format(stack)))
     return True
 
 

### CODE REVIEW

File: bootstrap_cfn/fab_tasks.py [hunk 1]
```diff
@@ -725,7 +731,15 @@ def cfn_create(test=False):
         # So delete the SSL cert that we uploaded
         if 'ssl' in cfn_config.data:
             iam.delete_ssl_certificate(cfn_config.ssl(), stack_name)
-        abort('Failed to create stack: {0}'.format(stack))
+        # delete the DNS tag
+        for elb in get_public_elbs():
```
ltsampros: I think a get_all_elbs() is more appropriate here.

bootstrap_cfn/fab_tasks.py
```diff
@@ -725,7 +731,15 @@ def cfn_create(test=False):
         # So delete the SSL cert that we uploaded
         if 'ssl' in cfn_config.data:
             iam.delete_ssl_certificate(cfn_config.ssl(), stack_name)
-        abort('Failed to create stack: {0}'.format(stack))
+        # delete the DNS tag
+        for elb in get_all_elbs():
```
yufangzhang: I created a PR to fix this chunk of code. basically if elb is empty, txt records won't be deleted. also txt record deletion is missing here. the PR has been merged. you could pull and have a look :p

bootstrap_cfn/fab_tasks.py
```diff
@@ -738,7 +748,14 @@ def cfn_create(test=False):
         # So delete the SSL cert that we uploaded
         if 'ssl' in cfn_config.data:
             iam.delete_ssl_certificate(cfn_config.ssl(), stack_name)
-        abort('Failed to create stack: {0}'.format(stack))
+        # delete the DNS tag
+        logger.info("Deleting DNS tag for '{stack} from '{zone}'...".format(stack=stack_id, zone=zone_name))
+        try:
+            print green("Removing tag stack from DNS...")
+            r53_conn.delete_txt_record(zone_name, zone_id, txt_tag_record)
```
yufangzhang, line 755: `r53_conn.delete_txt_record` only removes txt record. we probably need to delete/check Alias record too?

bootstrap_cfn/fab_tasks.py
```diff
@@ -738,7 +748,14 @@ def cfn_create(test=False):
...
+        except boto.route53.exception.DNSServerError:
+            print red("Failed to remove DNS tag")
+        abort(red('Failed to create stack: {0}. You may want to delete it manually.'.format(stack)))
```
yufangzhang, line 758: also, are we retaining the failed stack while removing the DNS records?

mattpep: Retaining the rolledback stack is current behaviour so I decided not to change it in this patch. I've a slight preference for *not* automatically deleting a rolledback stack but am open to discussion. But even if so, we should probably do it as a separate piece of work.

lukaszraczylo:
@mattpep & @yufangzhang any consensus on this or progress in merge?

sevenmachines:
@yufangzhang @mattpep @ltsampros Not quite sure why we're looking to delete records on rollback. We need them for cfn_delete the stack(?) Anyways, happy have my ignorance crushed by catching up and finding out :) But going to put in WIP for the moment

filipposc5:
A bit late to the party, but this PR is still open/pending. Niall's comment sounds right, if we are not dealing with a half created stack because we want to cfn_delete then we shouldn't need this PR altogether. So we should close this PR and document this expectation, perhaps under a workflow headline?

mattpep:
Invalid.

PR #261, opened 2017-04-04
cfn_update process for EC2 resources

ltsampros:
Partial support for cloudformation updates is now supported on the EC2 and ELB sections of the configuration file:

```fab application:courtfinder aws:my_project_prod environment:dev config:/path/to/courtfinder-dev.yaml tag:[tag_name] cfn_update```

Note: Running this command will show you some structured output of what
changes and how. Also a unified diff is printed on output between the
old and the new Launch Configuration sections. Although we have gone
to great lengths with this command, it can result in destructive
operations, particularly if one reduced the desired/max/min capacities
of the Auto Scaling Group.

Note #2: In order to test this, you will have to set PYTHONPATH pointing to this branch of the PR and to bootstrap-salt's version https://github.com/ministryofjustice/bootstrap-salt/pull/80

Note #3: The fabfile.py of the relevant -deploy repo should contain a line like this: ```from bootstrap_salt.fab_tasks import cfn_create, cfn_delete, cfn_update```

### DIFF
--- a/bootstrap_cfn/fab_tasks.py
+++ b/bootstrap_cfn/fab_tasks.py
@@ -717,7 +717,18 @@ def cfn_update(test=False):
     # Get online template
     response = cfn.conn_cfn.get_template(stack_name)
     body = response['GetTemplateResponse']['GetTemplateResult']['TemplateBody']
-    cfn.update(stack_name, cfn_config.process_update(body))
+    new_body = cfn_config.process_update(body)
+
+    x = raw_input("Are you sure you want to update the stack {}!? (y/n)\n".format(stack_name))
+    if x not in ['y', 'Y', 'Yes', 'yes']:
+        sys.exit(1)
+
+    rc = cfn.update(stack_name, cfn_config.process_update(body))
+    if not rc:
+        logger.critical("cfn_update: please check the logs for BotoServerError criticals")
+        logger.critical("cfn_update: this usually happens when cfn_update is ran but no changes are needed")
+        return
+
     tail(cfn, stack_name)
 
 @task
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: bootstrap_cfn/mime_packer.py [hunk 1]
```diff
@@ -46,7 +46,7 @@ def get_type(content, deftype):
 
 
 def pack(parts, opts={}):
-    outer = MIMEMultipart()
+    outer = MIMEMultipart(boundary='boostrapcfnboundary141')
```
ltsampros: I was thinking of using the library's version here but I think I might revert it to something more static. Versions are possible to change and the scripts to remain the same.

bootstrap_cfn/fab_tasks.py
```diff
@@ -687,6 +686,22 @@ def isactive():
...
+    Update the AWS cloudformation stack.
+    """
+    _validate_fabric_env()
+    stack_name = get_stack_name(new=False)
+    cfn_config = get_config(called_by_cfn_create=True)
+
+    cfn = get_connection(Cloudformation)
+    # Get online template
+    response = cfn.conn_cfn.get_template(stack_name)
+    body = response['GetTemplateResponse']['GetTemplateResult']['TemplateBody']
```
ltsampros, line 719: Some improvements/error checking etc are needed here.

yufangzhang (commented):
I tested it on adp-dev to resize instances and update ASG and ACM -- part of EC2 and ELB configuration. All changes applied. It may be better if we could highlight the diff changes a bit because currently it's a bit hard to tell.......
Also this code automatically updates docker (new user data etc) so it can break things..

ltsampros:
@yufangzhang This code is "updating docker" as a "side-effect". It needs an updated bootstrap-salt and currently the new b-salt version ships with a new urllib3 version. That version is required for new-docker but I don't think it breaks older docker formulas.

PR #262, opened 2017-04-19
Use canonical certificate name for lookup in template

yufangzhang:
we use canonical_certificate_name in template instead of the actual certificate_name


ltsampros:
LGTM.

ltsampros (approved):

PR #265, opened 2017-05-15
fixed bug in get_deployarn_record which resulted in jenkins deploy job failure

yufangzhang:
get_deployarn_record( ) returned `False`  that was not alerted in set_active_stack( ) and can break deploy jobs.

### DIFF
--- a/bootstrap_cfn/r53.py
+++ b/bootstrap_cfn/r53.py
@@ -178,7 +178,7 @@ class R53(object):
         for rr in rrsets:
             if rr.type == record_type and rr.name == record_name:
                 return rr.resource_records[0][1:-1]
-        return False
+        return None
 
     def get_full_record(self, zone_name, zone_id, record_name, record_type):
         """


bootstrap_cfn/elb.py
```diff
@@ -4,10 +4,10 @@
 
 import boto.ec2.elb
 
-from boto.exception import BotoServerError
-
 import boto.iam
 
+from boto.exception import BotoServerError
```
mattpep: Does this block of code need to be moved from its previous place?

ltsampros:
@yufangzhang I don't have a problem with the PR, however we have an inconsistency where the master_zone when defined doesn't need to end with a '.' while the hosted_zone on the elb section does.

I checked many of my checkout deployed repos and this is indeed the case. So, can we make sure that appending the dot doesn't break other places where this code is called?

yufangzhang:
I have split this PR into two to avoid confuses. this one now only fixes the bug we found in cait-deploy

filipposc5:
I see you removed the appending of the `.`, good call. We would need to test that more thoroughly before we jump to such behaviours. Looks good 👍

PR #267, opened 2017-05-17
Get hosted zone name and ID

yufangzhang:
use hosted_zone name and id for ELB

### DIFF
--- a/bootstrap_cfn/fab_tasks.py
+++ b/bootstrap_cfn/fab_tasks.py
@@ -481,8 +481,9 @@ def get_zone_name():
     return zone_name
 
 
-def get_zone_id():
-    zone_name = get_zone_name()
+def get_zone_id(zone_name=None):
+    if zone_name is None:
+        zone_name = get_zone_name()
     r53_conn = get_connection(R53)
     try:
         zone_id = r53_conn.get_hosted_zone_id(zone_name)
@@ -491,6 +492,18 @@ def get_zone_id():
     logger.info("fab_tasks::get_zone_id: Found zone id '%s' "
                 "for zone name '%s'...", zone_id, zone_name)
     return zone_id
+
+
+def get_elb_details(elb):
+    """
+    return (name, hosted_zone, hosted_id) of an elb
+    Args:
+        elb: (dict) elb object from yaml
+
+    Returns:
+        (tuple): elb name, hosted zone name, hosted zone id
+    """
+    return elb.get('name'), elb.get('hosted_zone'), get_zone_id(elb.get('hosted_zone'))
 
 
 def get_legacy_name():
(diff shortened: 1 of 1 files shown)


bootstrap_cfn/fab_tasks.py
```diff
@@ -1074,8 +1094,12 @@ def get_all_elbs(f=None):
     returns True, or everything.
     """
     cfn_config = get_config()
-    elbs = [x.get('name') for x in cfn_config.data.get('elb', {}) if x.get('scheme') in ['internet-facing', 'internal']]
-    return filter(f, elbs) if f else elbs
+    elbs = cfn_config.data.get('elb', {})
```
yufangzhang, line 1089: get_all_elbs() previously returns only elb['name'] list. This change makes it return the whole 'elb' object from yaml

filipposc5: `f` is not used anywhere.

filipposc5: if you don't need `f` skip the parameter. If you don't need to filter remove the parameter, from a quick search I could not find a valid place where the filter is used.

bootstrap_cfn/fab_tasks.py
```diff
@@ -492,6 +492,18 @@ def get_zone_id():
...
+    logger.info("fab_tasks::get_hosted_name: Found hosted zone name '%s' in config...",hosted_zone_name)
+    return hosted_zone_name
+
+
+def get_hosted_id(hosted_name):
+    r53_conn = get_connection(R53)
+    zone_id = r53_conn.get_hosted_zone_id(hosted_name)
+    logger.info("fab_tasks::get_hosted_id: Found hosted zone id '%s' in config...", zone_id)
+    return zone_id
+
```
filipposc5: What if [1]:
```
def get_elb_details(elb):
   return (elb['name'], elb['hosted_zone'], ....get_hosted_zone_id(elb['hosted_zone']))
```
and skip get_hosted_id since it is already implemented (rework method: `get_zone_id` to get a parameter like zone_name=None so it is compatible.

bootstrap_cfn/fab_tasks.py
```diff
@@ -624,10 +636,13 @@ def cfn_delete(force=False, pre_delete_callbacks=None):
             pass
 
     for elb in get_all_elbs():
-        logger.info("Deleting '{}-{}' from '{}' ({})...".format(elb, stack_id, zone_name, zone_id))
+        elb_name = elb.get('name', None)
+        hosted_name = get_hosted_name(elb)
+        hosted_id = get_hosted_id(hosted_name)
+        logger.info("Deleting '{}-{}' from '{}' ({})...".format(elb_name, stack_id, hosted_name, hosted_id))
```
filipposc5: and then
[2]: (elb_name, hosted_zone, hosted_id) = get_elb_details(elb)

bootstrap_cfn/fab_tasks.py
filipposc5: and if you must announce feel free to emit logger.info though I am not sure that it is needed.

filipposc5 (approved):

PR #268, opened 2017-05-19
Format zone name in r53 api calls

yufangzhang:


bootstrap_cfn/fab_tasks.py
```diff
@@ -275,13 +275,18 @@ def enter_maintenance(maintenance_ip, dry_run=False):
...
-        r53_conn.update_dns_record(zone_name, zone_id, elb, 'A', maintenance_ip, dry_run=dry_run)
+    try:
+        for elb in cfn_config.data['elb']:
+            if not apply_maintenance_criteria(elb):
+                continue
+            zone_id = get_cached_zone_id(r53_conn, cached_zone_ids, elb['hosted_zone'])
+            zone_name = get_zone_name()
+            print green("Attempting to update: \"{0}\":\"{1}\"".format(elb['name'], maintenance_ip))
+            r53_conn.update_dns_record(zone_name, zone_id, elb['name'], 'A', maintenance_ip, dry_run=dry_run)
+    except KeyError:
```
yufangzhang: here I only added try/except to catch KeyError and boto Error. the rest of code remains the same.

bootstrap_cfn/fab_tasks.py
```diff
@@ -322,35 +327,39 @@ def exit_maintenance(dry_run=False):
...
+                # alias_hosted_zone_id
+                R53.AWS_ELB_ZONE_ID[env.aws_region],
+                # alias_dns_name
+                dns_name,
+                # alias_evaluate_target_health (True/False)
+                False
+            ]
+            print green("Attempting to update: \"{0}\":{1}".format(elb['name'], record_value))
+            r53_conn.update_dns_record(get_zone_name(), zone_id, elb['name'], 'A', record_value, is_alias=True, dry_run=dry_run)
+    except KeyError:
```
yufangzhang: here I only added try/except to catch KeyError and boto Error. the rest of code remains the same

bootstrap_cfn/fab_tasks.py
filipposc5: These are already broken since they haven't been updated to work with multiple stacks so a bit pointless to update them, can you please remove them?

bootstrap_cfn/fab_tasks.py
filipposc5: These are already broken since they haven't been updated to work with multiple stacks so a bit pointless to update them, can you please remove them?

yufangzhang: @filipposc5 updated

bootstrap_cfn/r53.py
```diff
@@ -221,3 +229,6 @@ def get_all_resource_records(self, zone_id):
         """
         rrsets = self.conn_r53.get_all_rrsets(zone_id)
         return rrsets
+
+def consistent_zone_name(zone_name):
```
filipposc5: maybe rename to conform_with_fqdn()

yufangzhang:
how about now? @filipposc5 @ltsampros

filipposc5:
Apart from not being happy with the previous change in the r53 call, I think we can merge this! :)

PR #269, opened 2017-05-19
Update unittest

yufangzhang:
unittest fixes for support_old_bootstrap_cfn, get_all_elbs and update_dns_records


filipposc5:
Unfortunately we will have to break these tests again in the future but for now they are valid so gonna merge this and adjust later.

PR #272, opened 2017-06-06
Unblock tests

filipposc5:
One of the tests had a pdb settrace committed. Removing.


filipposc5:
This should fix coveralls reporting now...

PR #276, opened 2017-07-06
WIP: Allow logical name to be set on additional buckets

robyoung:
AWS allows buckets to have a logical name as well as a physical name. This feature is used by `upload_assets` to make it easier to perform actions on a kind of bucket within an environment. This is also useful for additional S3 buckets where we need a different bucket per environment.

Backwards compatibility is maintained by introducing an additional yaml key for the `logical-name` which overrides the resource name. If it's not set the resource name reverts to the bucket name.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -603,6 +603,17 @@ These entries can also specify their own policies or use the default, vpc limite
              /:
              expirationdays: 5
 
+A logical name can be provided for bucket which need not be unique. This can be useful when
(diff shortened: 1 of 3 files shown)


robyoung:
Apparently the same effect can be achieved by using the cloudformation output.

richerry (approved):
Approved

richerry (approved):
Approved

PR #277, opened 2017-07-10
Support alternative ACM domain domain validation domains

sevenmachines:
This change adds support for the domain_validation_options setting.
This means that any subject_alternative_names can have their
validation_domain explicitly overriden.


sevenmachines:
Passes unit tests but going to test on stack so WIP

sevenmachines:
tested on helloworld stack, looks ok

